PR #62, opened 2017-09-26
Auto rebuild docs

### LINKED ISSUE #55: Integrate generateFormatDocs with ReadTheDocs
Currently ReadTheDocs does not automatically generate the sources for the schema docs but we have to run make apidoc and check in the sources. This should be automated so that ReadTheDocs automatically generates the sources from the YAML specs directly.

oruebel:
This pull request integrates the generateFormatDocs.py script, which is used to generate Sphinx sources for the format documentation form the YAML specification files, with the Sphinx build directly. In this way the sources are automatically generated when we build the docs. The auto generated sources have therefore been removed from the repo.  Fix #55 . 

@ajtritt everything seems to work fine when I run it on my workstation. Is there an easy way to test if this works on ReadTheDocs before I do the merge?


oruebel:
Test on RTD fails due to the fact that the generateFormatDocs script needs FORM. I'll do some more work to see if I can get FORM installed on ReadTheDocs.

oruebel:
With the changes I've made the basic build is now working on ReadTheDocs, however, the rendering of the graphs still fails. Probably some other missing package. I'll make sure to fix this before I merge.

oruebel:
I've added the following main additional fixes to make the automatic build on ReadTheDocs work:

- Updated imports of other modules from our local docs/utils dir
- Added missing packages to the requirements conf
- Force use of Agg backend for matplotlib to avoid tkinter error
- Moved temp install of PyNWB for ReadTheDocs out of the doc source folder to avoid Sphinx warnings on RTD because sphinxs seems to pick up RST files it should not
- Increaded recursion depth for pickle in conf.py to avoid problem on ReadTheDocs
- Add support to check the git hash for which sources were autogenerated from YAML. This is to avoid redundant regeneration of sources to speed up the build and avoid timeouts on readthedocs
- Had to disable the epub build on readthedocs to avoid the timeouts. Even without generating the sources our builds were often close to the 900s time limit on ReadTheDocs and adding the additional work pushed the build frequently over the time limit.

Issue #50, opened 2017-08-31
Clarify format of `session_start_time` and `file_creation_date`


t-b:
The NWB spec says "Date + time, Use ISO format (eg, ISO 8601) or a format that is easy to read and unambiguous."

https://github.com/NeurodataWithoutBorders/nwb-schema/blob/4dfe947265fda43dd3e70f1cde098361b8451828/core/nwb.file.yaml#L11

https://github.com/NeurodataWithoutBorders/nwb-schema/blob/4dfe947265fda43dd3e70f1cde098361b8451828/core/nwb.file.yaml#L31)

This definition is too vague for interaction across different APIs.

I therefore propose to change the wording to`Strict ISO8601 format including timezone information and T separator:  “2017-08-28T23:24:47+02:00"`. Using a 64bit POSIX timestamp has the drawback that the reader needs to account for leap seconds, but would be equally fine for me.

Note: This issue is different than #49 as it does not change the specification but just clarifies it.

neuromusic:
+1

jonc125:
I'd prefer to go with the human readable ISO8601 format for this, as it's immediately interpretable by someone viewing the file with generic HDF5 tools, e.g. HdfView. If we want a POSIX timestamp, which I can see value of for software use, I think that should be a separate field, and possibly the new field proposed in #49.

neuromusic:
note @jonc125 that this issue (indeed, this entire repo) deals *only* with the schema and does not (necessarily) determine how the data is stored in HDF5. one of the major (and much needed) changes that @ajtritt & @oruebel have implemented in this refactoring of the NWB format is to separate the concerns of data storage, schema, and API

neuromusic:
in other words, even if we decide here that the schema should be in ISO8601, the backend (whether HDF5 or PostgreSQL or whatever) could still store the data as a POSIX timestamp.

similarly, even if we decide here that the schema should be in ISO8601, the API (pynwb) could still accept POSIX timestamps and coerce them to ISO8601

t-b:
@neuromusic But that would imply that we have a separate specification for the backend, or? Or how do you want otherwise exchange HDF5 files between NWB APIs without knowing how the file looks like?

neuromusic:
[...]

PR #71, opened 2017-10-06
NWB file session_start_time and file_create_date: Enforce ISO8601:2004(E) format

t-b:
To push #50 a bit further I decided to give it a shot via a PR.

### DIFF
--- a/core/nwb.file.yaml
+++ b/core/nwb.file.yaml
@@ -7,11 +7,13 @@ groups:
   datasets:
   - dims:
     - '*unlimited*'
-    doc: 'Time file was created, UTC, and subsequent modifications to file. COMMENT:
-      Date + time, Use ISO format (eg, ISO 8601) or a format that is easy to read
-      and unambiguous. File can be created after the experiment was run, so this may
-      differ from experiment start time. Each modifictation to file adds new entry
-      to array. '
+    doc: 'The point in time the file was created and subsequent modifications to the file.
+    COMMENT: The date and time must be given in ISO 8601:2004(E) format.
+      File can be created after the experiment was run, so this may differ from
+      experiment start time. Each modification to the file adds a new entry to
+      the array.
+      Only the following types of representations are allowed:
+      \"2017-08-28T23:24:47+02:00\", \"2017-08-30T5:21:47Z\" or \"2017-08-30T05:21:47.45925Z\".'
     dtype: text
     name: file_create_date
     shape:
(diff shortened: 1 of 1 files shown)


neuromusic:
Love it (though as I've said before, I'd prefer these fields to be "optional"). 

A few practical questions (mostly for @ajtritt)...

Since this isn't a native HDF5 type, how do we validate that files confirm? Does form need to check? If so, how does form know which fields need to be ISO8601?

ajtritt:
We would have to create a time format within FORM.  The two options that come to mind are

1) leave time as string, and FORM has to enforce the string fits the format on read/write. 
2) create a named type. This makes it more machine-readable

Either way, read APIs would be have the freedom to return the data in whatever format. For example, for PyNWB, it could be returned as a datetime object.

ajtritt:
Is there any disadvantage to taking @t-b's changes one step further, and requiring that time is always UTC?

neuromusic:
> Is there any disadvantage to taking @t-b's changes one step further, and requiring that time is always UTC?

Yes.  ISO8601:2004(E) is an international standard. It is widely used and was not defined by us, so we aren't unnecessarily reinventing the wheel. Using ISO8601:2004(E) ensures that these pieces of metadata are machine-readable (which is not currently guaranteed) & human-readable without over-constraining the NWB standard. Any library that can validate ISO8601 strings or coerce its own datetime objects into ISO8601 conforms without any further requirements.

t-b:
@ajtritt One disadvantage of always using UTC is also that one might argue that the timezone offset itself has information. 

With `named type` you mean a custom HDF5 data type? If yes, I would not recommend that as not all HDF5 APIs can read arbitraty custom datatypes (Java, Igor Pro, ...).

@neuromusic As I tried mentioning in the commit message we would allow only a subset of possible ISO8601 compliant date and time strings. This is done to make the correctness check easier.

ajtritt:
Okay, sounds like ISO8601:2004(E) is the way to go. 
Named types are just compound data types that get saved to the file for use throughout. I would create compound type that consists of all the fields in the ISO8601:2004(E) format, store it in file, and then the file_create_date and session_start_time datasets would link to this type. 

We already allow compound types, as it is the basis for tables.

t-b:
However we move forward on this issue. I suggest we at least teach pynwb to create timestamps in UTC timezone and not local timezone.

I just created a nwb file with latest `pynwb` and I get as timestamp "2018-04-03 14:16:28.158191" which is in PDT.

I can provide a PR of course.

oruebel:
Closing for now. The schema has changed significantly (due to refactoring) in PyNWB. These changes should be implemented in PyNWB and then we can sync the schema over to this repo afterwards. See also #50

PR #123, opened 2018-02-22
Enh/epoch refactor

### LINKED ISSUE #119: Restructure Epoch storage to make more efficient
# Problem

Right now Epochs are stored as a single group per Epoch. Within each Epoch, the index into each TimeSeries that the Epoch applies to is stored as a single group. This structure is inefficient for storing large numbers of Epochs. 

# Proposal

To make Epoch storage more efficient, we would like to make the following changes:

1) Replace the Epoch type with the new type EpochTable, a table with the following columns:
        - _start_ - the start time of each Epoch
        - _stop_ - the stop time of each Epoch
        - _description_ - a description for each Epoch
        - _tags_ - tags describing each Epoch
        - _ts_ref_ - a region reference into the TimeSeriesIndex (described below) for the TimeSeries indexed for each Epoch

2) Replace EpochTimeSeries with the new type TimeSeriesIndex table, a table with the following columns:
        - _start_idx_ - the start index in the TimeSeries
        - _count_ - the number of time points in the TimeSeries
        - _timeseries_ - an object reference to the TimeSeries object

ajtritt:
fix #119

### DIFF
--- a/core/nwb.base.yaml
+++ b/core/nwb.base.yaml
@@ -5,6 +5,27 @@ datasets:
     name: help
   doc: The attributes specified here are included in all interfaces.
   neurodata_type_def: NWBData
+- attributes:
+  - doc: Value is 'Data on how an epoch applies to a time series'
+    dtype: text
+    name: help
+    value: Data on how an epoch applies to a time series
+  doc: 'An index into a TimeSeries object'
+  dtype:
+  - doc: "Start index into the TimeSeries data[] field. COMMENT: This can be used\
+      \ to calculate location in TimeSeries timestamp[] field"
+    dtype: int32
+    name: idx_start
+  - doc: Number of data samples available in this time series, during this epoch.
+    dtype: int32
+    name: count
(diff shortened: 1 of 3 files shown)


bendichter:
I find it very hard to make sense of these yaml files. What we talked about makes sense to me. Maybe we should put this into another branch though and make changes to pynwb in another branch as well and then merge both when we are confident they are both working to everyone's satisfaction

oruebel (approved):
@ajtritt looks good to me. I would merge this once the corresponding pull in PyNWB passes all tests.

ajtritt:
@bendichter @neuromusic I added both of you so that you know what's going on. Don't worry about deciphering the YAML at the moment. Like @oruebel said, I will merge the PR when we make sure this works with the API.

bendichter:
sounds good!

bendichter (approved):

oruebel:
Changes will be pulled over when merging schema from PyNWB

PR #124, opened 2018-02-22
Enh/roi refactor

### LINKED ISSUE #118: Restructure ROI storage to make more efficient
# Problem
Right now ROIs are stored as a single group per ROI. This structure is inefficient for storing large numbers of ROIs. RoiResponseSeries' relationship to ROI objects is implicit (i.e. ROI is specified by a string), so must know a priori which ImageSegmentation and ImagingPlane was used to produce the ROIs.  

# Proposal
To make ROI storage more efficient and explicit,  we would like to make the following changes:

1. Replace ROI.img_mask with a 3D dataset with shape [num_rois, num_x_pixels, num_y_pixels] (i.e. an array of image masks)
2. Replace ROI.pix_mask/ROI.pix_mask_weight with a table that has columns “x”, “y”, and “weight” (i.e. ROI.pix_mask and ROI.pix_mask_weight into a single table)
3. Add an ROI table with columns "id", "description", "img_mask_ref", and "pix_mask_ref"
        - "id" holds a unique identifier for each ROI
        - "description" stores a description for each ROI
        - "img_mask_ref" stores a region reference into the img_mask dataset (described above in 1.)
        - "pix_mask_ref" stores a region reference into the pix_mask/pix_mask_weight dataset (described above in 2.)
4. RoiResponseSeries.roi_names gets replaced with RoiResponseSeries.rois, a scalar dataset that stores a region reference into the ROI table (described above in 3.)
5. Remove RoiResponseSeries.segmentation_interface. This information will be available through RoiResponseSeries.rois (described above in 4.)

ajtritt:
fix #118

### DIFF
--- a/core/nwb.ophys.yaml
+++ b/core/nwb.ophys.yaml
@@ -47,13 +47,11 @@ groups:
     shape:
     - null
     - null
-  - dims:
-    - num_ROIs
-    doc: List of ROIs represented, one name for each row of data[].
-    dtype: text
-    name: roi_names
-    shape:
-    - null
+  - doc: 'a dataset referencing into an ROITable containing information on the ROIs stored in this timeseries'
+    dtype:
+      target_type: ROITable
+      reftype: region
+    name: rois
   doc: ROI responses over an imaging plane. Each row in data[] should correspond to
     the signal from one ROI.
   links:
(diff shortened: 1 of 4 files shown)


core/nwb.ophys.yaml
```diff
@@ -121,46 +119,53 @@ groups:
...
+    - doc: a table for storing pixel masks
+      name: pixel_mask_weights
+      dtype:
+      - name: x
+        dtype: uint
+        doc: the pixel x-coordinate
+      - name: y
+        dtype: uint
+        doc: the pixel y-coordinate
+      - name: y
```
oruebel: should this be named ``weight`` instead of ``y``? I mean the dtype for pixel_mask_weights has the dimensions ``(x,y,y)`` instead of ``(x,y,weight)``

oruebel (changes requested):
Other than the name of the pixel_mask_weights this looks good to me. I would merge once it works in PyNWB.

core/nwb.ophys.yaml
ajtritt: you are correct

oruebel (dismissed):

oruebel (dismissed):

oruebel:
Changes will be pulled over when merging schema from PyNWB

PR #135, opened 2018-03-30
Nicify docs

t-b:
First try in getting the docs into shape which encourages enhancing them.

I'm still a bit fuzzy about the distinction between dimension and shape. Especially as the latter is null in 60 cases and only non-null in three cases.

### DIFF
--- a/core/nwb.base.yaml
+++ b/core/nwb.base.yaml
@@ -26,7 +26,7 @@ datasets:
     target_type: VectorData
     reftype: region
   shape:
-  - null
+  - 'null'
   neurodata_type_def: VectorIndex
   neurodata_type_inc: NWBData
 - default_name: element_id
@@ -38,7 +38,7 @@ datasets:
   doc: a unique identifier for each element
   dtype: int
   shape:
-  - null
+  - 'null'
   neurodata_type_def: ElementIdentifiers
   neurodata_type_inc: NWBData
 groups:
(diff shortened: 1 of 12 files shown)


oruebel:
@t-b thanks for the help. I'll have a look on Monday.

> I'm still a bit fuzzy about the distinction between dimension and shape.

The meaning and syntax of the dims an shape key are documented in the specification language docs (see links at the end). Basically, dims describes the names of dimensions and shape describes the shape of a dataset, i.e., the size of the different dimensions. Shape is null in most cases because in most cases the size of a datatset varies by experiment, however, in some case we do know beforehand how long a dimensions should be. E.g., in cases where a dimension is used to store physical (x,y,z) coordinates we know that the dimension should always have a length of 3. 

* http://schema-language.readthedocs.io/en/latest/specification_language_description.html#dims
* http://schema-language.readthedocs.io/en/latest/specification_language_description.html#shape

I hope this helps to clarify your question.

oruebel:
@t-b the schema has changed a lot in PyNWB (in part due to a refactor in  https://github.com/NeurodataWithoutBorders/pynwb/pull/601/ ) This makes it hard to merge your changes. I've ported your changes from this PR by hand to PyNWB in https://github.com/NeurodataWithoutBorders/pynwb/pull/616 and some the changes were also already part of  https://github.com/NeurodataWithoutBorders/pynwb/pull/601/ . I'll close this PR as it has been subsumed by the other ones on PyNWB. The changes will move over to this repo once we sync up the schema. @t-b sorry that I was not able to merge this PR, I appreciate your contributions!

PR #144, opened 2018-04-16
Removed content of docs/general and point to the new overview page

oruebel:
I ported all content of the docs/general over to the new overview page at https://neurodatawithoutborders.github.io/ . To avoid confusion and duplication, I've removed all main content from docs/general and am just pointing to the new page in case so that folks that may still use the old link know that they should visit the new page instead.


tarelli:
Looks great!

nicain:
DAMN!  Nice work!

Best,
-n

Issue #136, opened 2018-03-31
Change `ophys.ImagingPlane` fields `excitation_lambda` and `imaging_rate` from `str` to `float`


bendichter:
https://github.com/NeurodataWithoutBorders/nwb-schema/blob/a0042022f7cf1dc0c7a6a17bbfa17fc5c5e41824/core/nwb.ophys.yaml#L197

Is there a reason why these are stored as strings? It seems more sensible to store them as floats.

ajtritt:
This got carried over from 1.x, but I agree, a float or a double would make more sense.

elyall:
Also in all other classes, the sampling rate parameter is referred to as "rate". So I'd suggest changing the parameter's name ("imaging_rate" -> "rate") and the data type (str -> float).

elyall:
Alternatively, ImagingPlane's sampling rate parameter is redundant and can be removed altogether since all Series data tied to an ImagingPlane have to have their own rate/timestamps specified.

bendichter:
@oruebel @ajtritt how do you guys feel about removing imaging_rate from ImagingPlane as @elyall suggests? Might that cause issues I am not seeing?

oruebel:
This is fixed in https://github.com/NeurodataWithoutBorders/pynwb/pull/612

PR #155, opened 2018-04-28
ophys/imaging_rate: use float instead of text

jcfr:
See #136

### DIFF
--- a/core/nwb.ophys.yaml
+++ b/core/nwb.ophys.yaml
@@ -230,7 +230,7 @@ groups:
     dtype: float
     name: excitation_lambda
   - doc: Rate images are acquired, in Hz.
-    dtype: text
+    dtype: float
     name: imaging_rate
   - doc: Calcium indicator
     dtype: text


oruebel:
@jcfr did you update the schema and API class in PyNWB API for this too?

jcfr:
> did you update the schema and API in PyNWB API for this too?

No. 

As soon as the change are integrated in this repo, anyone will be able to run the pynwb script [./src/pynwb/data/UpdateFromUpstream.sh](https://github.com/NeurodataWithoutBorders/pynwb/pull/506/files#diff-415d73cb9f0aaa0ccd472cb6cc88bd14) to sync with the updated schema

(@oruebel  to enable this, would appreciate if you could review https://github.com/NeurodataWithoutBorders/pynwb/pull/506)

Now the update of the API would have to be done as part of the PR updating the schema in pynwb. May be @bendichter  could take care of this ?

oruebel:
@jcfr do I understand this correctly that the workflow here is to update the schema in nwb-schema first and to then catch up PyNWB to it? For consistency, it might be easier to do this the other way around, i.e. update the schema and API in PyNWB first and then push out the changes to nwb-schema.

jcfr:
> workflow here is to update the schema in nwb-schema first and to then catch up PyNWB to it?

Indeed.

> it might be easier to do this the other way around, i.e. update the schema and API in PyNWB first and then push out the changes to nwb-schema.

Definitively agree :+1:

oruebel:
The schema in PyNWB has changed quite a bit. I've ported and finished the changes in PyNWB (i.e., updated the container class and added tests). The changes will move over the this repo when we copy the schema from PyNWB to here. Closing this PR. The current PR for this is on the PyNWB repo here https://github.com/NeurodataWithoutBorders/pynwb/pull/612

PR #198, opened 2018-10-30
add boolean to dtypes table

bendichter:
yoked to https://github.com/NeurodataWithoutBorders/pynwb/pull/691


oruebel:
@bendichter can you also update the dtype table in the storage docs accordingly https://github.com/NeurodataWithoutBorders/nwb-schema/blob/dev/docs/storage/source/storage_hdf5.rst#dtype-mappings

oruebel:
Changes added in #213

Issue #194, opened 2018-10-09
add electrodes as optional argument to ecephys.Clustering


bendichter:
It would be great if `ecephys.Clustering` could support an optional `electrodes` argument that would expect an electrode_table_region object.

tjd2002:
If you use the Unit table for unit metadata instead of Clustering, then you can just add a column for electrodes (actually an electrodeTableRegion). For instance, some of our Unit columns look like this:

```python
    # External clustering software gives names for each cluster--we want to preserve these
    nwbf.add_unit_column('cluster_name',  '(str) cluster name from clustering software')
    nwbf.add_unit_column('elec_group',    '(electrodeGroup) nTrode on which spikes were recorded')

    # For tetrode data, this will usually be all channels in the tetrode
    nwbf.add_unit_column('neighborhood',  '(electrodeTableRegion) list of electrodes on which spikes were clustered')
```

Notably, the datatypes (str, electrodeTableRegion,...) are not enforced by the schema in this approach, but that seems like something that could be addressed separately.

tjd2002:
Same logic applies to #111, which requests new fields be added to Clustering.

bendichter:
[...]

Issue #239, opened 2018-10-16
combine SpikeEventSeries and Clustering


bendichter:
## 2) Feature Request

How would people feel about folding `Clustering` into `SpikeEventSeries` by removing `Clustering` and adding a `SpikeEventSeries.cluster` field of dtype `int`? We could make `data` optional, in case a user only wants to store the times and not the waveforms (this is pretty common).

This would:
1) Make the relationship between a spike time and the spike waveform more explicit
2) Remove redundant `timestamps` that is currently duplicated across the two objects
3) Resolve the issue that `Clustering` should have an `electrodes` field (https://github.com/NeurodataWithoutBorders/nwb-schema/issues/194)
4) Resolve the issue that `Clustering` should be a `TimeSeries`(https://github.com/NeurodataWithoutBorders/nwb-schema/issues/112)

## Checklist

- [x] Have you ensured the feature or change was not already [reported](https://github.com/NeurodataWithoutBorders/pynwb/issues/) ?
- [x] Have you included a brief and descriptive title?
- [x] Have you included a clear description of the problem you are trying to solve?
- [x] Have you included a minimal code snippet that reproduces the issue you are encountering?

ajtritt:
The only thing I am unsure of is making `SpikeEventSeries.data` optional. If a SpikeEventSeries doesn't have `data`, then it fails to meet the requirements of a `TimeSeries`.

tjd2002:
I'm all for removing 'Clustering'. I've been thinking of Clustering+SpikeEventSeries as being superseded by the new Unit table (for Unit metadata including clustering metrics) and UnitTimes (for spike times of each clustered unit), and expecting that all that older stuff would soon be deprecated. We've been using Unit + UnitTimes to store the output of clustering.

I think this use of Unit + UnitTimes handles the case where there is no per-spike waveform data. So we could keep SpikeEventSeries and add the '.cluster' field as proposed, and then only use it when storing waveforms. I think this would mean we don't have to worry about the no-data case discussed above.

tjd2002:
Conversation seems to be continuing over at #194, in case anyone else is following along.

stephprince:
Closing because `Clustering` has been deprecated.

PR #245, opened 2019-01-17
[WIP] add UnitSeries neurodata_type

ajtritt:
This is a proposal/work-in-progress for addressing #194. It adds a new type, called *UnitSeries*, that would replace *Clustering*. It extends *TimeSeries* and requires that *TimeSeries.data* be an `int`. It also adds an additional dataset, called *units* that is a DynamicTableRegion that points to *Units* table.


tjd2002:
Would this allow two or more spikes from different units with the same timestamp? (This can happen in real data)

ajtritt:
@tjd2002 

> Would this allow two or more spikes from different units with the same timestamp? (This can happen in real data)

Yes, there are no restrictions on unique time stamps in *TimeSeries*

bendichter:
I'm fine with the name change, and I like that this now inherits from `TimeSeries`, but it would be nice if there was some kind of link to `SpikeEventSeries` (https://github.com/NeurodataWithoutBorders/nwb-schema/issues/239). As it stands now, if you had multiple `UnitSeries` and multiple `SpikeEventSeries` objects (e.g., 6, one for each of 6 shanks), it is unclear how you would indicate which `SpikeEventSeries` object is connected to which `UnitSeries` object. I guess you could do it implicitly if both had a field for referencing the regions of the `Units` table, and you implied that objects that referenced the same Units table region were connected.

rly:
I believe this is covered by the Units table with its waveforms column and link to the electrodes table. Please re-open if not.

PR #256, opened 2019-04-03
Added missing docs for how format specifications are chached

oruebel:
The storage docs where missing the description of how the format specification is cached in HDF5.


t-b:
+1

PR #265, opened 2019-05-22
update author list

bendichter:
I don't know precisely what the order should be but I think this is an improvement

### DIFF
--- a/core/nwb.namespace.yaml
+++ b/core/nwb.namespace.yaml
@@ -2,15 +2,19 @@ namespaces:
 - name: core
   doc: NWB namespace
   author:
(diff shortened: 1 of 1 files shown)


oruebel:
Let's change this in PyNWB first and then migrate it over here since we have a few additional changes (e.g,. SpecFile and DecompositionSeries) that we need to migrate over as well.

rly:
@bendichter I believe this is the same change as #266 and so has been merged. Close?

rly:
Oh. @bendichter did you intend to do PR #266 on pynwb and not nwb-schema? the namespace YAML in pynwb hasn't been changed yet.

PR #285, opened 2019-07-13
Document object_id

### LINKED ISSUE #284: Describe ObjectId
https://github.com/hdmf-dev/hdmf/pull/90/ adds an ObjectId to all neurodata_types. This should be documented in the documentation of the HDF5 backend (and any other backends).

rly:
Resolves #284


docs/format/source/format_description.rst
```diff
@@ -203,6 +203,22 @@ Extending *NWBContainer* works in the same way, e.g., to create more specific ty
 data processing.
 
 
+Common attributes
+-----------------
+
+All NWB:N Groups and Datasets have three required attributes: `neurodata_type`, `namespace`, and `object_id`.
```
oruebel: Instead of "All NWB:N Groups and Datasets" I think we should say something like "All NWB:N Groups and Datasets with an assigned neurodata_type have ...."

docs/storage/source/storage_hdf5.rst
```diff
@@ -102,14 +103,15 @@ Datasets
     linkable                      Not mapped; Stored in schema only
```
oruebel, line 103: I don't think ``linkable`` is part of the schema language any more, I think this is left-over from NWB 1.

rly: It looks like it is still in the schema language. Just not used. https://schema-language.readthedocs.io/en/latest/specification_language_description.html?highlight=linkable#groups

oruebel (approved):

PR #290, opened 2019-08-14
A few minor fixes/enhancements

yarikoptic:
Please see individual commit messages

PS just found these on my hard drive not submitted, rebased, seems still apply

### DIFF
--- a/bin/find_missing_help.py
+++ b/bin/find_missing_help.py
@@ -28,7 +28,7 @@ def find_missing_help(d):
 for curr_file in sys.argv[1:]:
     with open(curr_file) as fin:
         fd = yaml.safe_load(fin)
-        for d in fd['groups']:
+        for d in fd.get('groups', []):
             find_missing_help(d)
 
 if len(missing) > 0:
--- a/bin/find_missing_help.py
+++ b/bin/find_missing_help.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
 import ruamel.yaml as yaml
 import sys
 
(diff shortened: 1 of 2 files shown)


bin/find_missing_help.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
```
ajtritt, line 1: Is this Windows friendly?

bin/reformat_spec.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
```
ajtritt, line 1: Is this Windows friendly?

ajtritt (changes requested):

bin/reformat_spec.py
yarikoptic: I think Windows doesn't care about shebangs... SO seems to confirm: https://stackoverflow.com/questions/7574453/shebang-notation-python-scripts-on-windows-and-linux
so this would be of help on OSX/*nix and of no (positive/negative) effect on windows

bin/find_missing_help.py
yarikoptic: very ;)

bin/reformat_spec.py
ajtritt: Why is this one `python3` and the other one `python`? Shouldn't the environment dictate the version?

yarikoptic: Eh, forgot detail. May be because this one isn't python 2 compatible. I am ok to change to regular python here

ajtritt (approved):

PR #302, opened 2019-08-22
Add Events and LabeledEvents types

### LINKED ISSUE #301: Add Events type
It is often useful to store events as a list of timestamps, without packaging them into a TimeSeries. For example, this list could represent when the animal receives reward, when the animal performs a particular behavior, etc. This is different from TimeIntervals because there is no stop time; these are instantaneous events. Also, TimeIntervals has the baggage of being a DynamicTable. This is simpler than AnnotationSeries which has all the baggage of being a TimeSeries.

@bendichter and I propose two new types for nwb.event.yaml:
- an Events object for simple events (e.g. when the animal receives reward)
- a LabeledEvents object for events that can take on multiple values (e.g. if there are multiple reward categories, the reward category can be encoded alongside the event time)
  - enumerated types would work great here, but are a pain to implement, so we propose a mapping type, where there is a 1-D dataset of integer labels (length equal to number of timestamps, value from 0 to number of labels - 1), and a 1-D dataset for the mapping of text (length equal to number of labels), such that the i-th element corresponds to a label value of i (zero-indexed)

rly:
Fixes #301

### DIFF
--- a/core/nwb.event.yaml
+++ b/core/nwb.event.yaml
@@ -15,7 +15,6 @@ groups:
     - null
     doc: Event timestamps, in seconds, relative to the common experiment master-clock stored in 
       NWBFile.timestamps_reference_time.
-    quantity: '?'
     attributes:
     - name: unit
       dtype: text
--- a/core/nwb.event.yaml
+++ b/core/nwb.event.yaml
@@ -8,7 +8,7 @@ groups:
     doc: Description of the event.
   datasets:
   - name: timestamps
-    dtype: float
+    dtype: float32
     dims:
     - num_events
     shape:
@@ -34,7 +34,7 @@ groups:
     labels[100] is 'Full reward', and all other entries in 'labels' are the empty string.
   datasets:
   - name: label_keys
-    dtype: uint
+    dtype: uint8
     dims:
     - num_events
     shape:
--- a/core/nwb.event.yaml
+++ b/core/nwb.event.yaml
@@ -8,7 +8,7 @@ groups:
     doc: Description of the event.
   datasets:
   - name: timestamps
-    dtype: float64
+    dtype: float
     dims:
     - num_events
     shape:
--- a/core/nwb.event.yaml
+++ b/core/nwb.event.yaml
@@ -25,7 +25,7 @@ groups:
   neurodata_type_inc: Events
   doc: A list of timestamps, stored in seconds, of an event that can have different labels. For example,
     this type could represent the times that reward was given, as well as which of three different 
-    types of reward was given. In this case, the 'labels' dataset would contain values {0, 1, 2},
+    types of reward was given. In this case, the 'label_keys' dataset would contain values {0, 1, 2},
     and the 'label_map' would contain three text elements, where the first (index 0) specifies the
     name of the reward associated with a label = 0, the second (index 1) specifies the name of the
     reward associated with a label = 1, etc. The labels do not have to start at 0 and do not need to be
--- a/core/nwb.event.yaml
+++ b/core/nwb.event.yaml
@@ -42,7 +42,7 @@ groups:
     doc: Integer labels that map onto strings using the mapping in the 'label_map' dataset. Values must
       be 0 or greater and need not be sequential. This dataset should have the same number of elements 
       as the 'timestamps' dataset.
-  - name: label_map
+  - name: labels
     dtype: text
     dims:
     - num_values
--- a/core/nwb.event.yaml
+++ b/core/nwb.event.yaml
@@ -26,20 +26,20 @@ groups:
   doc: A list of timestamps, stored in seconds, of an event that can have different labels. For example,
     this type could represent the times that reward was given, as well as which of three different 
     types of reward was given. In this case, the 'label_keys' dataset would contain values {0, 1, 2},
-    and the 'label_map' would contain three text elements, where the first (index 0) specifies the
-    name of the reward associated with a label = 0, the second (index 1) specifies the name of the
-    reward associated with a label = 1, etc. The labels do not have to start at 0 and do not need to be
-    sequential, e.g. the 'labels' dataset could contain values {0, 10, 100}, and the 'label_map' dataset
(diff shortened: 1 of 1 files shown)


core/nwb.event.yaml
```diff
@@ -0,0 +1,53 @@
...
+- neurodata_type_def: Events
+  neurodata_type_inc: NWBDataInterface
+  doc: A list of timestamps, stored in seconds, of an event
+  attributes:
+  - name: description
+    dtype: text
+    doc: Description of the event.
+  datasets:
+  - name: timestamps
+    dtype: float64
```
bendichter:
```suggestion
    dtype: float
```
I prefer `float`. This will allow the user to use `float32` or `float64`.

core/nwb.event.yaml
```diff
@@ -0,0 +1,53 @@
...
+    dims:
+    - num_times
```
bendichter:
```suggestion
    - num_events
```

core/nwb.event.yaml
```diff
@@ -0,0 +1,53 @@
...
+    this type could represent the times that reward was given, as well as which of three different 
+    types of reward was given. In this case, the 'labels' dataset would contain values {0, 1, 2},
+    and the 'label_map' would contain three text elements, where the first (index 0) specifies the
+    name of the reward associated with a label = 0, the second (index 1) specifies the name of the
+    reward associated with a label = 1, etc. The labels do not have to start at 0 and do not need to be
+    sequential, e.g. the 'labels' dataset could contain values {0, 10, 100}, and the 'label_map' dataset
+    could contain 101 values, where label_map[0] is 'No reward', label_map[10] is '10% reward', 
+    label_map[100] is 'Full reward', and all other entries in 'label_map' are the empty string.
+  datasets:
+  - name: labels
```
bendichter:
```suggestion
  - name: label_keys
```

core/nwb.event.yaml
```diff
@@ -0,0 +1,53 @@
...
+    dtype: int
+    dims:
+    - num_times
+    shape:
+    - null
+    doc: Integer labels that map onto strings using the mapping in the 'label_map' dataset. Values must
+      be 0 or greater and need not be sequential. This dataset should have the same number of elements 
+      as the 'timestamps' dataset.
+  - name: label_map
```
bendichter:
```suggestion
  - name: labels
```

core/nwb.event.yaml
bendichter: How would you feel about making this dataset optional? If it is not supplied, the label keys are `0, 1, 2, ...` Another option would be to keep this as is and build that into the API.

core/nwb.event.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - name: unit
+      dtype: text
+      value: seconds
+      doc: Unit of measurement for timestamps, which is fixed to 'seconds'.
+      
+- neurodata_type_def: LabeledEvents
+  neurodata_type_inc: Events
+  doc: A list of timestamps, stored in seconds, of an event that can have different labels. For example,
+    this type could represent the times that reward was given, as well as which of three different 
+    types of reward was given. In this case, the 'labels' dataset would contain values {0, 1, 2},
```
bendichter:
```suggestion
    types of reward was given. In this case, the 'label_keys' dataset would contain values {0, 1, 2},
```

bendichter (changes requested):

core/nwb.event.yaml
ajtritt: Might be a good use of DynamicTable. Users could then add additional metadata to their labels.

core/nwb.event.yaml
rly: In the spec, I'm pretty sure `float` corresponds to `float32`. See https://schema-language.readthedocs.io/en/latest/specification_language_description.html#dtype and the corresponding code in https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/spec/spec.py#L22-L42 .

The timestamps of `TimeSeries` has dtype `float64` and I think timestamps should have higher precision in general.

bendichter: The spec defines the *minimum allowable precision* so `float32` would allow you to use `float64`

rly: Ah, I see that. Sorry, I missed that. What do you think about replacing `float64` with `float` for the other instances below?
TimeSeries starting_time
TimeSeries timestamps
SpikeEventSeries timestamps
FeatureExtraction times
EventDetection times

There are also 16 places where int32 and two places where uint64 is specified, most if not all of which do not require such high precision.

bendichter: I'm in favor of flexible precision throughout. I can't see a downside

bendichter (dismissed):

bendichter (dismissed):

bendichter:
I am putting together an extension [here](https://github.com/ben-dichter-consulting/ndx-point-cloud-table) that would benefit from having `Events` as a `DynamicTable`

core/nwb.event.yaml
```diff
@@ -0,0 +1,53 @@
...
+    types of reward was given. In this case, the 'label_keys' dataset would contain values {0, 1, 2},
+    and the 'labels' dataset would contain three text elements, where the first (index 0) specifies the
+    name of the reward associated with a label_keys = 0, the second (index 1) specifies the name of the
+    reward associated with a label_keys = 1, etc. The labels do not have to start at 0 and do not need to
+    be sequential, e.g. the 'label_keys' dataset could contain values {0, 10, 100}, and the 'labels' 
+    dataset could contain 101 values, where labels[0] is 'No reward', labels[10] is '10% reward', 
+    labels[100] is 'Full reward', and all other entries in 'labels' are the empty string.
+  datasets:
+  - name: label_keys
+    dtype: int
```
bendichter: If "Values must be 0 or greater, " shouldn't this be `uint`?

rly: Now changed.

rly:
The proposed `Events` and `LabeledEvents` types will be implemented in an NDX extension in development here: https://github.com/rly/ndx-events

PR #308, opened 2019-09-14
Clarify documentation of external image storage in ImageSeries

oruebel:


core/nwb.image.yaml
```diff
@@ -87,18 +88,22 @@ groups:
...
-    doc: Format of image. If this is 'external' then the field external_file contains
-      the path or URL information to that file. For tiff, png, jpg, etc, the binary
-      representation of the image is stored in data. If the format is raw then the
-      fields bit_per_pixel and dimension are used. For raw images, only a single channel
-      is stored (e.g., red).
-    quantity: '?'
+    doc: Format of image. If this is 'external', then the attribute 'external_file'
+      contains the path information to the image files. If this is 'raw', then the raw
+      (single-channel) binary data is stored in the 'data' dataset. If this attribute
+      is not present, then then the default format='raw' case is assumed.
```
rly: Should we add `default_value: 'raw'`?

core/nwb.image.yaml
```diff
@@ -38,8 +38,9 @@ groups:
   neurodata_type_inc: TimeSeries
   doc: General image data that is common between acquisition and stimulus time series.
     Sometimes the image data is stored in the file in a raw format while other
-    times it will be stored as an external image file in the host file system. The
-    data field will either be binary data or empty. [frame][y][x] or [frame][z][y][x].
+    times it will be stored as a series of external image files in the host file system.
```
bendichter: It seems like the original description had something like tif stacks in mind, which is a common way to store acquired data and saves all of the images in a single file, like a video.

core/nwb.image.yaml
oruebel: Adding a `default_value: 'raw'`  is fine, but we should not make the field required in order to not break backward compatibility.

rly (approved):

oruebel:
@nclack this PR should clarify the use of external image stacks with ImageSeries. The structure of ImageSeries itself has not changed, only the documentation has been refined and clarified.

PR #312, opened 2019-09-16
Allow flexible precision -- int32 -> int, float32, float64, double -> float

### LINKED ISSUE #311: Allow flexible precision for all types
re discussion with @bendichter in https://github.com/NeurodataWithoutBorders/nwb-schema/pull/302/files

Problem: many parts of the specification require high precision. If the data are actually collected at lower precision, this higher precision storage is both a waste of space and misleading to the actual precision. There are also many inconsistencies over which data require higher precision and which don't. 

Suggestion: the [schema documentation](https://schema-language.readthedocs.io/en/latest/specification_language_description.html#dtype) says "The precision indicated in the specification is generally interpreted as a minimum precision. Higher precisions may be used if required by the particular data."

What do you think about using the minimum precision for all types. i.e. change all int32 -> int, all {float32, float64, double} -> float? See the specific changes in my associated PR.

rly:
Fix #311


bendichter:
@oruebel @ajtritt this wouldn't change anything, right? I think these are good changes. It is easier to interpret the schema this way.

rly:
I forgot that int32 == int and float32 == float. The PR changes those words, which hopefully indicates that the value can be higher precision, but maybe it is better to make the minimum number of bits is more explicit (and therefore change all "int" -> "int32" and "float" -> "float32")

oruebel:
For PyNWB I don't think this would change anything. I think the main folks affected by this change would be tool developers. Having the precision fixed in the schema makes things easier in typed languages (e.g., C/C++). I don't think it is an issue for backward compatibility as this relaxes constraints, rather than adding.

oruebel:
> and therefore change all "int" -> "int32" and "float" -> "float32"

That sounds good as it is more explicit (without actually changing semantics)

ajtritt:
No, this doesn't change anything. 

short = int16
int = int32
long = int64
float = float32
double = float64

I think the only thing we're missing is "half = float16"

t-b:
@rly I'll have a look at this PR this week. (I'm currently travelling so please don't merge before ;))

rly:
Relatedly, I found this (perhaps outdated) text on precision for time values and in general: https://nwb-schema.readthedocs.io/en/latest/format_description.html#storing-time-values

If we change the above, we should change this text as well. 

Also it is worth mentioning that AFAIK, most, if not all, ephys acquisition systems store timestamps as **integers** with at least 32 bits. We store raw time series data as ints with a conversion factor in order to preserve the data in its rawest form. For the same reason, we could also store raw timestamps as ints with a conversion factor.

bendichter:
@rly That's a good point about time. In fact, we do have a time conversion factor in the schema, it's just always set to `1`. If we instead have it default to 1 but flexibly set in the API, we could add this feature without breaking anything.

oruebel:
> Relatedly, I found this (perhaps outdated)

The text is outdated at least in the sense that we specify minimum precision only and no longer support "!" as part of data types. 

> If we change the above, we should change this text as well.

Yes that text should be updated.

>  For the same reason, we could also store raw timestamps as ints with a conversion factor.

I think leaving float64 for the timestamps in TimeSeries is fine for now. The timestamps are not the bulck of the data and storing them explicitly (rather than requiring a conversion) should be fine. I'd wait at least until we have a concrete ask for this.

t-b:
Looks good. I could not spot anymore strictly specified types with `git grep -B 3 -nHP "((uint|int|float)(32|64)|double)" core/`.

Can we loosen the requirements for text as well? Currently we have `ascii` in the following places

```
$git grep -B 3 -nHP ascii                                                                                            
nwb.base.yaml-279-    of usability.                                                                                            
nwb.base.yaml-280-  attributes:                                                                                                                  
nwb.base.yaml-281-  - name: colnames                                                                                                                 nwb.base.yaml:282:    dtype: ascii                                                                                                     
--                                                                                                                                     
nwb.file.yaml-323-          doc: Impedance of the channel.                                                                             
nwb.file.yaml-324-        - name: location                                                                                      
nwb.file.yaml-325-          neurodata_type_inc: VectorData                                                                                          
nwb.file.yaml:326:          dtype: ascii                                                                                                             
--                                                                                                                     
nwb.file.yaml-339-          doc: Reference to the ElectrodeGroup this electrode is a part of.
nwb.file.yaml-340-        - name: group_name
nwb.file.yaml-341-          neurodata_type_inc: VectorData
nwb.file.yaml:342:          dtype: ascii
```

This is one of the open issues for me with NWBv2 files from Igor Pro as I can only write UTF8 encoded text.

Unfortunately with these changes existing files created with d7f23a1a (Repo cleanup (#1055) 2019-09-17) dont' validate anymore. See the following log
[log.txt](https://github.com/NeurodataWithoutBorders/nwb-schema/files/3640019/log.txt).

t-b:
@rly @oruebel My comment in https://github.com/NeurodataWithoutBorders/nwb-schema/pull/312#issuecomment-533902982 regarding the fact that old files don't validate anymore has been resolved as we decided that only NWB files with attached specification must validate with the latest pynwb version or?

@rly Could you rebase?

rly:
The updated PR changes "int" -> "int32", "uint" -> "uint32", "float" -> "float32", and "double" -> "float64" to be explicit about the data type precision. Since the new terms are synonymous with the old terms, this does not functionally change the schema and does not break backwards compatibility.

ajtritt (approved):

PR #318, opened 2019-10-08
BF: do not include NWB- prefix in an example for nwb_version

yarikoptic:
Also made it 2.1.0 to match current one in core/nwb.file.yaml

### DIFF
--- a/orig/schema.json
+++ b/orig/schema.json
@@ -1958,7 +1958,7 @@
                         "data_type": "text"
                     },
                     "nwb_version": {
-                        "description": "File version string. COMMENT: Eg, NWB-1.0.0. This will be the name of the format with trailing major, minor and patch numbers.",
+                        "description": "File version string. COMMENT: Eg, 2.1.0. This will be the name of the format with trailing major, minor and patch numbers.",
                         "data_type": "text"
                     }
                 },


rly:
As far as I know, `orig/schema.json` is kept here for legacy/historical reasons. I think it is a YAML schema for 1.0.4_beta, which is not even the latest 1.x version (see [1.0.6 pdf specification](https://github.com/NeurodataWithoutBorders/specification_nwbn_1_0_x/tree/master/version_1.0.6)) 

@oruebel what purpose does this file serve and should we be updating it? 

@yarikoptic For the latest schema for NWB 2.x, see the YAML files in the `core` directory

oruebel:
> what purpose does this file serve and should we be updating it?

The file is purely legacy. It was used as a starting point when we created the 2.0 schema. We should not be updating this. In 1.0 the schema was hard-coded in a Python dict https://github.com/NeurodataWithoutBorders/api-python/blob/master/nwb/nwb_core.py and we needed a reference to develop from.

yarikoptic:
thank you for the quick reply!
I will try to remember that when I `git grep` for terms, I should ignore that file (hint: may be worth just removing it and leaving legacy behind ;))

oruebel:
> may be worth just removing it and leaving legacy behind

We could just move it to the https://github.com/NeurodataWithoutBorders/specification_nwbn_1_0_x
repo

PR #305, opened 2019-08-30
Deprecate manifold in favor of pixel_origin and pixel_deltas

rly:
Implementing spec changes from https://github.com/NeurodataWithoutBorders/pynwb/pull/576

PR #322, opened 2019-11-04
Update documentation for ImagingPlane.reference_frame

rly:
https://github.com/NeurodataWithoutBorders/nwb-schema/pull/305 deprecated `manifold` of `ImagingPlane` but the documentation for the `reference_frame` dataset still refers to `manifold`. This PR updates that documentation to refer to the new datasets instead of `manifold`.

### DIFF
--- a/core/nwb.ophys.yaml
+++ b/core/nwb.ophys.yaml
@@ -246,7 +246,7 @@ groups:
     - 2
     - 3
     doc: Space between pixels in (x, y) or voxels in (x, y, z) directions, in the specified unit.
-      Assumes imaging plane is a regular grid.
+      Assumes imaging plane is a regular grid. See also reference_frame to interpret the grid.
     required: false
     attributes:
     - name: unit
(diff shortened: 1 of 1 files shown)


oruebel:
@lydiang

oruebel (dismissed):

rly:
I updated the documentation further and removed the "unit" attribute of "origin_coords" because the value of that dataset should be the coordinates in pixel (or voxel) space of (0,0) in some real-world reference frame. Thus, there are no units. @bendichter 

I imagine these datasets will be further refined and probably replaced as we integrate ontologies for describing anatomical coordinates, but for now this should suffice. @lydiang

rly:
@bendichter @oruebel can you review this and check that my interpretation of how these variables: origin_coords, grid_spacing, and reference_frame makes sense?

core/nwb.ophys.yaml
```diff
@@ -234,13 +234,9 @@ groups:
     shape:
     - 2
     - 3
-    doc: Location of the origin in (x, y) or (x, y, z), in the specified unit.
+    doc: Pixel location of the origin in (x, y) or (x, y, z), zero-indexed. See also reference_frame to interpret the
```
oruebel: If I understand this correctly, this changes the ``origin_coords`` from being an absolute location in ``meters`` in physical space to an (x,y,z) location in image space. Not sure if that makes sense, as it is unclear what you are translating the image relative to. Also ``origin_coords`` is a float, which would imply sub-pixel translations?

rly: To be clear, I think we did not fully think through `origin_coords` during its design. The image is being translated relative to whatever coordinate space is described in `reference_frame`. 

Your comment makes me think it would be better to specify the location in meters (based on some reference frame) of just the first element (0,0) or (0,0,0) of the matrix. Or maybe that is what we were thinking initially? That way it is more straightforward to describe an imaging plane that does not include the origin of the reference frame.

oruebel (approved):

PR #323, opened 2019-11-04
Add waveform sampling rate and unit to Units table

### LINKED ISSUE #273: units/waveform_mean and units/waveform_sd missing sampling rate
Maybe I am missing something but I do not think it is possible to determine the sampling rate of `waveform_mean` and `waveform_sd` in the `Units` table. To solve this, we could add an optional `waveform_sampling_rate` attribute to the `Units` table that is required if these columns are present.

rly:
Fix #273

### DIFF
--- a/core/nwb.misc.yaml
+++ b/core/nwb.misc.yaml
@@ -175,6 +175,16 @@ groups:
   default_name: Units
   doc: Data about spiking units. Event times of observed units (e.g. cell, synapse,
     etc.) should be concatenated and stored in spike_times.
+  attributes:
+  - name: waveform_sampling_rate
+    dtype: float
+    doc: Sampling rate of waveforms stored in waveform_mean and waveform_sd, in hertz.
+    required: false
+  - name: waveform_unit
+    dtype: text
+    value: volts
+    doc: Unit of measurement for waveforms stored in waveform_mean and waveform_sd. This value is fixed to 'volts'.
+    required: false
   datasets:
   - name: spike_times_index
     neurodata_type_inc: VectorIndex
--- a/core/nwb.misc.yaml
+++ b/core/nwb.misc.yaml
@@ -175,16 +175,6 @@ groups:
   default_name: Units
   doc: Data about spiking units. Event times of observed units (e.g. cell, synapse,
     etc.) should be concatenated and stored in spike_times.
-  attributes:
-  - name: waveform_sampling_rate
-    dtype: float
-    doc: Sampling rate of waveforms stored in waveform_mean and waveform_sd, in hertz.
-    required: false
-  - name: waveform_unit
-    dtype: text
-    value: volts
-    doc: Unit of measurement for waveforms stored in waveform_mean and waveform_sd. This value is fixed to 'volts'.
-    required: false
   datasets:
   - name: spike_times_index
     neurodata_type_inc: VectorIndex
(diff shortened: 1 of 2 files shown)


rly:
I would recommend making both attributes `sampling_rate` and `unit` required in a future major release, but for now, they are both optional in order to maintain backwards compatibility.

oruebel (changes requested):
Since this is on the schema repo, we should also update the release notes in conjunction with this change. Other than that this looks good.

rly:
@oruebel good idea. I have updated those in this PR.

oruebel (approved):

bendichter:
This would solve the waveform smoking rate issue, but there are a few different spikes processing systems that require the acquisition sampling rate even when waveforms aren't present. Is there a way to address that?

rly:
What are those systems using the acquisition sampling rate for when
waveforms are not present? If it is not meaningfully used, can they just
use nan or another placeholder value when it is not found in the NWB file?

oruebel:
> but there are a few different spikes processing systems

Could you clarify which systems need this and why they need the sampling rate if no waveforms are available?

bendichter:
@oruebel Specifically, SpikeInterface, expects spike times as `uint64`, corresponding to the indexes of the samples where they occurred. Spikes, which is used by the Giocomo Lab has a similar constraint. In both cases, we need to supply the sampling rate manually when the acquisition is not present, because it is not discernible from the file otherwise. In general, it might be useful to know the temporal resolution with which spikes were found.

I think this PR is a good solution to the waveform sampling rate problem and I don't want to hold this up because of an issue that is related but not the same. @rly have you thought about what the syntax might be for inputing this information in pynwb?

oruebel:
> expects spike times as `uint64`, corresponding to the indexes of the samples where they occurred

If I understand this correctly, then this means SpikeInterfaces reads the ``spike_times``  from the ``units`` table in the NWB file---which are floating point times in second relative to the time stamp reference time---and transforms them into a corresponding integer index, effectively describing at which index the spike would be stored if we had a sequence of regularly sampled timestamps. Since this is specific to the ``spike_times`` I think (similar to what this PR is doing for the waveforms) you'd want an attribute on the ``spike_times`` column that stores the precision of the time values (which is determined by the sampling rate of acquisition.) There are two factors that make the spike times different though: 1) spike times are points in times rather than a regularly sampled signal like waveforms and 2) depending on the method used to determine spike times, the spike time may have higher precision than the acquisition (i.e., what I mean is the spike time may be determined to be between acquired timestamps). As such, I think rather than ``sampling_rate`` it may be more appropriate to call this ``time_precision``. Would we need the same attribute also for the ``obs_intervals`` column?

bendichter:
In reality SpikeInterface and Spikes both want the sampling rate of the acquisition, and assume that this is sufficient temporal precision (which is a good assumption). If you want to call this `time_precision` to avoid this being confused with a time series that has data at regular temporal intervals, that makes sense to me.

Observation intervals are generally determined by things like whether or not the amp was plugged in, or movement of a probe, so I don't think the same temporal precision would apply there. All the tools I know of that would use this information would be happy to work with time in seconds, and would not need to convert to samples. If they are working with the converted `int64` spike time values, they will need to apply the exact same conversion as spike times, so they can use the attribute from the `spike_times` array.

bendichter (approved):

oruebel:
@bendichter adding a ``precision`` attribute on the ``spike_times`` column to address this use-case seems reasonable to me.

Issue #258, opened 2019-04-23
Add metadata to Device


rly:
To add information about a recording device, e.g. who was the manufacturer, what kind of amplifier, what version is it, what settings were used (or analogous information about an imaging microscope), then I would have to put that into the `name` field or into fields of the NWBFile like `data_collection`. It would be more intuitive to include this metadata information as part of the Device object, which is pretty empty - it has only the `name` field. We could bundle this information in a catch-all `description` field or we could structure this metadata information with specific fields, such as `manufacturer`, `hardware_version`, etc.

Currently, in PyNWB:
```python
device = nwbfile.create_device(name='Plexon Omniplex')
```
Suggestion:
```python
device = nwbfile.create_device(name='Omniplex', 
                               description='Plexon Omniplex A with Analog Amplifier, software version 1.14.0')
```
OR
```python
device = nwbfile.create_device(name='Omniplex',
                               manufacturer='Plexon',
                               hardware_version='Omniplex A with Analog Amplifier',
                               software_version='1.14.0')
```

bendichter:
As long as these additional fields are optional it won't be schema-breaking.

serial_number might also be a good field

t-b:
Does it really make sense to add 1+X new ones instead of just an optional description? Both solutions don't help in machine reading so using just description would help equally good.

bendichter:
@t-b you don't think someone might want to query for a session with a specific software version?

t-b:
@bendichter The format of the additional strings is fully user defined. So my reasoning is that we only need one these. Does that make sense?

ajtritt:
> The format of the additional strings is fully user defined

@t-b on that same reasoning, any data that is stored as a string is a user-defined format. Until a CV/ontology has been designed for these fields, they won't be optimally useful.

However, In the interest of progress over perfection, I think we should lay out a few common fields that users can populate as they see fit.

t-b:
@ajtritt 

> I think we should lay out a few common fields that users can populate as they see fit.

Sounds good, I'm not tied to any particular solution.

oruebel:
@rly it sounds like you and others will be looking at vendor formats during the hackathon. It might make sense to have a look at what the intersection of common fields are in the vendor formats and add optional fields that are common across vendors. I.e., I would assume that (even though specific name keys will be different) that many vendors will have some set of common metadata they store. 

[...]

PR #326, opened 2019-11-13
Add description and manufacturer to Device, extract to new file

rly:
Partially handles #258 but we should consider eventually add more structured fields to Device aside from just `manufacturer`.

### DIFF
--- a/core/nwb.file.yaml
+++ b/core/nwb.file.yaml
@@ -258,10 +258,9 @@ groups:
         ADC boards, microscopes, etc.
       quantity: '?'
       groups:
-      - neurodata_type_def: Device
-        neurodata_type_inc: NWBContainer
-        doc: A data acquisition device, e.g. recording system.
-        quantity: '+'
+      - neurodata_type_inc: Device
+        doc: Data acquisition devices.
+        quantity: '*'
     - neurodata_type_def: Subject
       neurodata_type_inc: NWBContainer
       name: subject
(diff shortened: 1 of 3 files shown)


bendichter (approved):

core/nwb.file.yaml
```diff
@@ -258,10 +258,9 @@ groups:
         ADC boards, microscopes, etc.
       quantity: '?'
       groups:
-      - neurodata_type_def: Device
-        neurodata_type_inc: NWBContainer
-        doc: A data acquisition device, e.g. recording system.
-        quantity: '+'
+      - neurodata_type_inc: Device
+        doc: Data acquisition devices.
+        quantity: '*'
```
rly, line 263: Was there a reason why Device had quantity '+' (one or more)? I guess it makes sense that an NWB file should have some data and that data should have some device information attached to it. But then it means our barebones examples of making an NWBFile results in an invalid file (in practice, the validator says such a file is valid). 

I lean toward saying that an NWB file can be valid without any data and without any device.

bendichter: I agree

oruebel: I agree

bendichter:
@rly there's a problem here that I did not catch at first. Attributes cannot have the property "quantity". These two properties should instead say `required: False`. This is causing a problem when matnwb tried to auto-generate these classes because it does not read the quantity attribute and instead makes these attributes required, which is the default behavior.

rly:
@bendichter Good catch. Thanks!

PR #330, opened 2019-11-18
add optional rel_x,y,z as relative positions

### LINKED ISSUE #325: electrode position
Right now the x,y,z columns of the electrodes table are ambiguous and under-utilized. The problem is that it's not clear what space they are supposed to belong to. The most common I've seen across datasets is to store the x,y position of an electrode on a probe. This is used by e.g. SpikeExtractors, kilosort, etc. to determine which electrodes are close enough to each other to potentially observe the same neuron. The other possible use is brain-space e.g. the Allen CCF. Eddie Chang stores x,y,z in brain space as well. This is less common to know, but obviously just as useful. I think we really need spots for both. The problem is right now we only have one set of x,y,z, and it isn't clear to a user which way it is intended to be used. Maybe we could add some optional columns like `group_x`, `group_y` that would only be valid within a group relative to the other electrodes in the group, or `device_x`, `device_y` that would be valid within a device.

bendichter:
fix #325

### DIFF
--- a/core/nwb.ecephys.yaml
+++ b/core/nwb.ecephys.yaml
@@ -232,11 +232,16 @@ groups:
       of area/layer, etc. Use standard atlas names for anatomical regions when possible.
   datasets:
   - name: position
-    dtype: float
-    shape:
-    - 3
-    dims:
-    - x, y, z
+    dtype:
+    - name: x
+      dtype: float
+      doc: x coordinate
+    - name: y
+      dtype: float
+      doc: y coordinate
+    - name: z
+      dtype: float
+      doc: z coordinate
     doc: stereotaxic or common framework coordinates
     quantity: '?'
   links:
--- a/core/nwb.ecephys.yaml
+++ b/core/nwb.ecephys.yaml
@@ -229,8 +229,16 @@ groups:
   - name: location
     dtype: text
     doc: Location of electrode group. Specify the area, layer, comments on estimation
-      of area/layer, stereotaxic coordinates if in vivo, etc. Use standard atlas
-      names for anatomical regions when possible.
+      of area/layer, etc. Use standard atlas names for anatomical regions when possible.
+  datasets:
+  - name: position
+    dtype: float
+    shape:
+    - 3
+    dims:
+    - x, y, z
+    doc: stereotaxic or common framework coordinates
+    quantity: '?'
   links:
   - name: device
     target_type: Device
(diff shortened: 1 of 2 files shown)


core/nwb.ecephys.yaml
```diff
@@ -229,8 +229,16 @@ groups:
...
-      of area/layer, stereotaxic coordinates if in vivo, etc. Use standard atlas
-      names for anatomical regions when possible.
+      of area/layer, etc. Use standard atlas names for anatomical regions when possible.
+  datasets:
+  - name: position
+    dtype: float
+    shape:
+    - 3
+    dims:
+    - x, y, z
```
rly: Rather than having a 1-d array of length 3, I would use either three scalar datasets for x, y, and z (like in the electrodes table, for consistency), or a compound data type for x, y, and z, which is what we wished we had done for the electrodes table.
```yaml
dtype:
- name: x
  dtype: float
  doc: x coordinate
- name: y
  dtype: float
  doc: y coordinate
- name: z
  dtype: float
  doc: z coordinate
```

core/nwb.file.yaml
```diff
@@ -349,6 +349,21 @@ groups:
           neurodata_type_inc: VectorData
           dtype: ascii
           doc: Name of the ElectrodeGroup this electrode is a part of.
+        - name: rel_x
+          neurodata_type_inc: VectorData
+          dtype: float
+          doc: x coordinate in channel group
```
rly:
```suggestion
          doc: x coordinate in electrode group
```

core/nwb.file.yaml
```diff
@@ -349,6 +349,21 @@ groups:
...
+          quantity: '?'
+        - name: rel_y
+          neurodata_type_inc: VectorData
+          dtype: float
+          doc: y coordinate in channel group
```
rly:
```suggestion
          doc: y coordinate in electrode group
```

core/nwb.file.yaml
```diff
@@ -349,6 +349,21 @@ groups:
...
+          quantity: '?'
+        - name: rel_z
+          neurodata_type_inc: VectorData
+          dtype: float
+          doc: z coordinate in channel group
```
rly:
```suggestion
          doc: z coordinate in electrode group
```

rly (changes requested):

rly (approved):

PR #337, opened 2019-11-27
Define ElectrodesTable and FrequencyBandsTable types

### LINKED ISSUE #336: Electrodes table and frequency bands table should be their own types
The "electrodes" table in `NWBFile` and the "bands" table in `DecompositionSeries` inherit from `DynamicTable` and add required datasets (columns) to them, but they themselves are not their own types. This makes validation problematic both at runtime and for the IO object.

Proposal: add a `neurodata_type_def` to these tables: `Electrodes` and `FrequencyBands`

See https://github.com/NeurodataWithoutBorders/pynwb/issues/1115

rly:
Fix #336


oruebel (dismissed):

bendichter:
This won't cause validation or IO errors for files that already exist? If this can be done without disrupting users I think it's a good idea

oruebel:
I think on the PyNWB the validation and read of existing files can be handled. @bendichter does this change anything for MatNWB?

bendichter:
@oruebel I don't know, possibly. I can check

bendichter:
I'm having trouble getting the matnwb to work. They are several schema changes behind at this point.

bendichter:
This break `matnwb/tutorials/ecephys.m` because the electrodes table is created manually as a `DynamicTable`, which needs to change to an `Electrodes` object. I'll fix that and see if there are any other issues

bendichter:
I would prefer to have the convention that tables are called Tables, i.e. this would be an `ElectrodesTable` object, similar to how we name descendants of `TimeSeries` with Series

bendichter:
@rly status update: this is working on the matnwb side

rly:
> I would prefer to have the convention that tables are called Tables, i.e. this would be an `ElectrodesTable` object, similar to how we name descendants of `TimeSeries` with Series

I would prefer this convention as well. However, we already have the `Units` table and the `TimeIntervals` table (which can have fixed names "epochs", "trials", "invalid_times"). The icephys extension will also add a few more `DynamicTable` types: `IntracellularRecordings`, `SweepSequences`, `Runs`, and `Conditions`.

bendichter:
I was talking about a convention for the neurodata_type name, not the object name, so it's not an issue for "epochs", "trials", "invalid_times". I agree we have types that break this convention, but I think we should use it going forward. At least when I create extensions I am going to try to use this naming convention for children of DynamicTable. At any rate, this schema change looks good to me besides that.

rly:
I renamed the types and suggested a renaming of the icephys extension types: https://github.com/oruebel/ndx-icephys-meta/issues/27

If we go forward with this convention, I also suggest renaming `Units` to `UnitsTable` and making `Units` an alias to `UnitsTable` to maintain backwards compatibility. Same for `TimeIntervals` to `TimeIntervalsTable`.

bendichter:
@rly that sounds scary. I'm worried about backwards compatibility, esp for matnwb

rly:
Superseded by #539 and #610

PR #342, opened 2019-12-17
core/nwb.icephys.yaml: Avoid requiring 64bit integers

t-b:
The author introduced 64bit unsigned integers with the introduction of
the sweep table in pynwb with 21f91920 (Add a SweepTable for icephys
data, 2018-11-02).

As it turns out this is not a clever thing to do. For one the types in
the schema are always minimum types so if we ever had 64bit
sweep numbers, this would be schema compliant as well.

And some software packages, like Igor Pro 8, can not write 64bit
integers.

So let's only require it to be uint.


oruebel:
If this is only reverting an unpublished changed, then this is fine. Otherwise, can you add a line to the 2.2 release notes here so we don't forget to mention it https://github.com/NeurodataWithoutBorders/nwb-schema/blob/dev/docs/format/source/format_release_notes.rst

oruebel (dismissed):

t-b:
@oruebel I've added something to the release notes.

oruebel (dismissed):

PR #346, opened 2019-12-28
fix RoiResponseSeries doc

bendichter:


core/nwb.ophys.yaml
```diff
@@ -32,9 +32,8 @@ groups:
 
 - neurodata_type_def: RoiResponseSeries
   neurodata_type_inc: TimeSeries
-  doc: ROI responses over an imaging plane. Each row in data should correspond to
-    the signal from one ROI. Each element on the second dimension of data should
-    correspond to the signal from one ROI.
+  doc: ROI responses over an imaging plane. Each column in data should correspond to
+    the signal from one ROI.
```
rly: I would prefer using "first dimension" and "second dimension" over "rows" and "columns". How "rows" and "columns" are mapped to dimensions/axes can be unclear. 

Suggested change:
doc: ROI responses over an imaging plane. The first dimension represents time. 
  The second dimension, if present, represents ROIs.

rly (changes requested):

core/nwb.ophys.yaml
bendichter: sounds good

rly (approved):

PR #356, opened 2019-12-29
fix ImageSeries.fied_of_view

### LINKED ISSUE #349: TwoPhotonSeries.field_of_view schema error
https://github.com/NeurodataWithoutBorders/nwb-schema/blob/76fc024c39815e5d3dfb62d958f9c451fa973570/core/nwb.ophys.yaml#L16-L27

The "required" field is [not allowed](https://schema-language.readthedocs.io/en/latest/specification_language_description.html#sec-dataset-spec) for datasets.

bendichter:


oruebel:
@bendichter I think it would be simpler to put updates of release notes in the same PR as the change of the schema. In this way we could easily confirm that the release notes match the change in the schema and we avoid issues where one PR has been merged but the other has not.

oruebel (dismissed):

bendichter:
@oruebel Yes, I think I am doing that. Same PRs but different commits, so that might be a bit confusing.

PR #366, opened 2020-01-01
Create pythonapp.yml

bendichter:


bendichter:
This is working now! It is failing because there are some schema language errors that this test correctly finds. I have open PRs for most if not all of them. It uses a specific commit of pynwb, which can be relaxed as soon as the validation code is merged.

bendichter:
close in favor of #379

PR #369, opened 2020-01-02
set Image dtype to numeric

### LINKED ISSUE #368: Specify dtype for Image and GrayscaleImage
see #347

bendichter:
set Image dtype to numeric
fix #368

### DIFF
--- a/core/nwb.base.yaml
+++ b/core/nwb.base.yaml
@@ -5,6 +5,7 @@ datasets:
 
 - neurodata_type_def: Image
   neurodata_type_inc: NWBData
+  dtype: numeric
   dims:
   - - num_x
     - num_y
--- a/core/nwb.image.yaml
+++ b/core/nwb.image.yaml
@@ -8,6 +8,7 @@ datasets:
   - null
   - null
   doc: A grayscale image.
+  dtype: numeric
 
 - neurodata_type_def: RGBImage
   neurodata_type_inc: Image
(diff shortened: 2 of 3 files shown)


oruebel (dismissed):

rly (approved):

rly:
@bendichter as an aside, is it necessary for subclasses of Image to also define `dtype`? Shouldn't they inherit `dtype` from Image? Or are you just being explicit about the dtype?

bendichter:
@rly How exactly we want to treat dtype is still under discussion [here](https://github.com/NeurodataWithoutBorders/nwb-schema/pull/370). I realized recently that `dtype` was defined by the schema language as a required key, but is often omitted. Ttrying to add `dtype` will cause problems for non-specific classes like `VectorData` that could be numeric or string, so we decided to make it optional specifically in those types of cases. You are right that for specific classes we could have `dtype` be inherited from a parent, but I thought we might as well be explicit if we can and stay as close as we can to the original intent of the schema language. Do you think it would be better to remove dtypes that could be inherited? I'd be open to that.

rly:
@bendichter I assume the dtype is automatically inherited by subclasses if it is not specified. Is that not the case? 

I am fine with both being explicit about the dtype in a subclass and omitting an inherited dtype to remove redundancy and simplify the schema, though in general, I have a slight preference for the latter. So, this is good with me.

PR #370, opened 2020-01-02
make dtype optional

### LINKED ISSUE #367: Make dtype optional for datasets
see #347

bendichter:
and explain that it is to be specified for concrete classes

fix #367

### DIFF
--- a/docs/language/source/specification_language_description.rst
+++ b/docs/language/source/specification_language_description.rst
@@ -850,7 +850,7 @@ The specification of a datasets is described in YAML as follows:
         neurodata_type_inc: Optional neurodata_type the group should inherit from
         quantity: Optional quantity identifier for the group (default=1).
         linkable: Boolean indicating whether the group is linkable (default=True)
-        dtype: Required string describing the data type of the dataset
+        dtype: Optional string describing the data type of the dataset
         dims: Optional list describing the names of the dimensions of the dataset
         shape: Optional list describing the shape (or possible shapes) of the dataset
         value: Optional to fix value of dataset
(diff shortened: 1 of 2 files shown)


docs/language/source/specification_language_description.rst
```diff
@@ -850,7 +850,7 @@ The specification of a datasets is described in YAML as follows:
         neurodata_type_inc: Optional neurodata_type the group should inherit from
         quantity: Optional quantity identifier for the group (default=1).
         linkable: Boolean indicating whether the group is linkable (default=True)
-        dtype: Required string describing the data type of the dataset
+        dtype: Optional string describing the data type of the dataset
```
oruebel, line 853: We should add "(default=null)" to clarify that when dtype is not specified that this is the same as ``dtype: null``.

docs/language/source/specification_language_description.rst
```diff
@@ -921,7 +921,7 @@ Boolean describing whether the this group can be linked.
 ``dtype``
 ^^^^^^^^^
 
-String describing the data type of the dataset. Same as for attributes. See :numref:`sec-dtype` for details.
+String describing the data type of the dataset. Same as for attributes. See :numref:`sec-dtype` for details. ``dtype`` may be omitted for abstract classes. Best practice is to define ``dtype`` for most concrete classes.
```
oruebel, line 925: Minor clarification for the last two sentences. 

"``dtype`` may be omitted or set to ``null`` for abstract classes, indicating that the dataset or attribute may take on any valid data type. Best practice is to define ``dtype`` for concrete classes."

bendichter:
@oruebel As far as I can tell,`dtype: null` is not currently used anywhere in the NWB schema. Why do you want to add this to the language?

oruebel:
I don't think this is adding to the language, but as far as I can tell, this is the default behavior (at least in PyNWB/HDMF) when dtype is omitted. I guess it is Ok not mention it, since we don't use this behavior.

oruebel (approved):

rly (approved):

PR #373, opened 2020-01-02
set the stim desc for IZeroCurrentClamp

### LINKED ISSUE #372: IZeroCurrentSeries stimulus description
By definition, an `IZeroCurrentSeries` has no stimulus, so it should not allow for a "stimulus description"

cross-listed from https://github.com/NeurodataWithoutBorders/pynwb/issues/1130

bendichter:
fix #372

An IZeroClampSeries has no stimulus, so this attribute is automatically set to "NA"

This is technically schema-breaking, if someone used a description other than the default "NA" for an `IZeroCurrentClampSeries`.


t-b:
I think this is a good change, even if schema-breaking, as it fixes a bug.

ajtritt (dismissed):

core/nwb.icephys.yaml
```diff
@@ -69,6 +69,11 @@ groups:
     and amplifier settings are off (i.e., CurrentClampSeries fields will be zero).
     There is no CurrentClampStimulusSeries associated with an IZero series because
     the amplifier is disconnected and no stimulus can reach the cell.
+  attributes:
+  - name: stimulus_description
+    dtype: text
+    doc: An IZeroClampSeries has no stimulus, so this attribute is automatically set to "NA"
+    value: NA
```
rly:
```suggestion
    doc: An IZeroClampSeries has no stimulus, so this attribute is automatically set to "N/A"
    value: N/A
```

rly: It is more common to see Not Applicable abbreviated as "N/A" or "n/a" then "NA"

rly (approved):

PR #379, opened 2020-01-03
add CI test

bendichter:
* add compound type support
* add uint type
* add titles


nwb.schema.json
```diff
@@ -51,6 +53,7 @@
             "int32",
             "int16",
             "int8",
+            "uint",
```
rly, line 100: uint64 is missing

rly: "utf-8" is also misspelled below

rly: Technically, short, long, bytes, datetime, and datetime64 are also allowed. See HDMF https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/spec/spec.py#L23-L41 . The specification language docs do not mention these data types.

.github/workflows/pythonapp.yml
```diff
@@ -0,0 +1,22 @@
...
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+    - name: Install pynwb
+      run: |
+        pip install git+https://github.com/NeurodataWithoutBorders/pynwb.git@cd3c352087a159dfab635c0cde4239fbede0973a
```
rly: Does this line mean this file needs to update every time pynwb updates (or the validate_spec method of pynwb updates)? Perhaps the spec validator should live in this repo.

rly (changes requested):

.github/workflows/pythonapp.yml
bendichter: I would like the spec validator to be importable, which means it would need to live in a repo that contains a setup.py and is released on pypi. For that reason I put it in pynwb, but it might be better suited for hdmf. I included a specific commit because the validator spec has not yet been merged into pynwb, so a simple pip install will not install it.

bendichter:
see https://github.com/hdmf-dev/hdmf/pull/230 for removal of dtype `datetime64`

bendichter:
Let's put this on hold until we move the validation code to hdmf

.github/workflows/pythonapp.yml
```diff
@@ -0,0 +1,22 @@
...
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+    - name: Install pynwb
+      run: |
+        pip install git+https://github.com/hdmf-dev/hdmf.git
```
bendichter, line 18: I'll change this to the released version once we release the recently added validator running code

rly (approved):

PR #396, opened 2020-01-17
set dims and shape of OpticalSeries.data for RGB images

### LINKED ISSUE #316: OpticalSeries does not allow for RGB
The `dims` of `OpticalSeries` inherits from `ImageSeries`, which is:

```yaml
- - frame
   - x
   - y
- - frame
   - x
   - y
   - z
```
This makes it impossible to properly record the presentation of colored images. I would expect this to be the most common application of the neurodata_type. There are ways to present volumetric data to he subject, but presenting color images is much more common, and I am not aware of any groups using `OpticalSeries` to record the presentation of volumetric stimuli.

My preferred solution would be to change the shape of `OpticalSeries.data` to:

```yaml
- - frame
   - x
   - y
- - frame
   - x
   - y
   - r,g,b
```
This issue is currently blocking the Rutishauser lab, which is trying to write and publish a guide for converting data to NWB.

bendichter:
fix #316

### DIFF
--- a/core/nwb.image.yaml
+++ b/core/nwb.image.yaml
@@ -146,6 +146,7 @@ groups:
     doc: Width, height and depth of image, or imaged area, in meters.
     quantity: '?'
   - name: data
+      doc: Images presented to subject, either grayscale or RGB
       dtype: numeric
       dims:
         - - frame
--- a/core/nwb.image.yaml
+++ b/core/nwb.image.yaml
@@ -145,6 +145,24 @@ groups:
     - - 3
     doc: Width, height and depth of image, or imaged area, in meters.
     quantity: '?'
+  - name: data
+      dtype: numeric
+      dims:
+        - - frame
+          - x
+          - y
+        - - frame
+          - x
+          - y
+          - r, g, b
+      shape:
+        - - null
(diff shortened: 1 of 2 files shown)


oruebel (dismissed):

rly (dismissed):

rly (dismissed):

rly:
As this is a bugfix, I think this is fine to include in version 2.2.2 rather than 2.3.0.

bendichter:
This requires an associated change on a pynwb branch, right?

bendichter:
the doc is missing. This should not validate. Can we include the json-schema tests in the CI?

rly:
@bendichter yes, this requires an associated change in PyNWB. I am working on the enh/update_schema branch for that.

The doc is missing, but isn't it inherited from `ImageSeries.data`?

rly:
(To be sure, I would add a doc key anyway)

rly (approved):

rly:
@bendichter please merge if this is ready to go. I have included your associated API change within PR https://github.com/NeurodataWithoutBorders/pynwb/pull/1146

PR #399, opened 2020-01-18
Reorder requirements-doc.txt to fix pynwb-hdmf dependency problem

### LINKED ISSUE #397: Release notes are not up-do-date
See https://nwb-schema.readthedocs.io/en/latest/format_release_notes.html.

rly:
By pip installing PyNWB before hdmf-docutils, the pinned version of HDMF that is compatible with PyNWB should be installed instead of the latest incompatible version of HDMF (installed by hdmf-docutils).

Fixes #397.

### DIFF
--- a/requirements-doc.txt
+++ b/requirements-doc.txt
@@ -1,2 +1,2 @@
-hdmf-docutils
 pynwb
+hdmf-docutils


rly:
Apparently one of the changes in the recent HDMF successfully fixed a compatibility issue that prevented the docs from building, so this PR is no longer necessary.

PR #400, opened 2020-01-20
rmv wrong change

bendichter:
This change was never made and should not be made. Not sure how it got added to this list, but it should be removed.


rly:
That line comes from https://github.com/NeurodataWithoutBorders/nwb-schema/pull/354/files and it looks like the schema change wasn't reverted. Perhaps the text should be changed to "Changed "quantity: '?'" of attribute Units.resolution to "required: false" for schema language compliance."?

rly:
@bendichter Could you please review this PR?

PR #408, opened 2020-02-06
Add new type FocalDepthImage, clean up retinotopy.yaml

rly:
After discussion with @ajtritt , it is recommended not to add a required attribute (`focal_depth`) to an instance of neurodata_type (`RetinotopyImage`) without creating a new neurodata type. To get around that, here we create a new neurodata type to store the focal depth image.


rly:
This change will be deferred to a future version.

PR #422, opened 2020-03-19
Create release notes for version 2.2.3 and set version to 2.2.3a

rly:
This adds release notes / changelog missing that was missing in PR #421.

### DIFF
--- a/core/nwb.file.yaml
+++ b/core/nwb.file.yaml
@@ -7,7 +7,7 @@ groups:
   attributes:
   - name: nwb_version
     dtype: text
-    value: 2.2.3-alpha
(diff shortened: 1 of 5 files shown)


rly:
I also updated the version from `2.2.2` to `2.2.3-alpha` for consistency across the schema and better tracking in case an API accidentally uses the schema in the dev branch. Let me know if you would prefer a different scheme, e.g., one that updates with every commit to dev, such as `2.2.3-a0` for the first change, `2.2.3-a1` for the next change, etc. That might be overdoing it though.

rly:
Closing since most of these changes were made in a separate PR and version 2.2.3 has been released.

PR #424, opened 2020-03-27
Add software process document

rly:
This PR proposes a process for making commits and releases to this repo and adds it to the documentation.

### DIFF
--- a/docs/format/source/software_process.rst
+++ b/docs/format/source/software_process.rst
@@ -6,23 +6,28 @@ language specification, and storage specification. PRs should be made to "dev". 
 the corresponding format/language/storage release notes. If the PR involves a change to the NWB format, the PR should
 also update the version of the format in three places: `/docs/format/conf.py`, `/core/nwb.namespace.yaml` and
 `/core/nwb.file.yaml`. The new version string should be the next bugfix/minor/major version of the format with the
-suffix "-alpha". For example, if the format is currently in version "2.2.0" and the format is then updated with a bug
-fix, then the PR for that bug fix should update the format version to "2.2.1-alpha". Appending the "-alpha" tag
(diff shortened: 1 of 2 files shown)


docs/format/source/software_process.rst
```diff
@@ -0,0 +1,28 @@
+The "dev" Branch and Releases
+=============================
+
+The default branch is "dev". The "dev" branch will hold the bleeding edge version of the NWB format specification,
+language specification, and storage specification. PRs should be made to "dev". Every PR should include an update to
+the corresponding format/language/storage release notes. If the PR involves a change to the NWB format, the PR should
+also update the version of the format in three places: `/docs/format/conf.py`, `/core/nwb.namespace.yaml` and
+`/core/nwb.file.yaml`. The new version string should be the next bugfix/minor/major version of the format with the
+suffix "-alpha". For example, if the format is currently in version "2.2.0" and the format is then updated with a bug
```
oruebel: This is inconsistent with https://nwb-extensions.github.io/versioning_guidelines  I think we should just use the ``a``, ``b``, and ``rc`` version tags from the versionining guidelines rather than ``-alpha``.

bendichter (commented):
This is great. The only thing I would add is that it is also important that `hdmf` and `pynwb` always link to released versions of the schemas as well. I know that's really the responsibility of those repos, but it is very important so I think it's worth mentioning here anyway.

rly:
@bendichter I added a couple sentences about that. Let me know what you think.

docs/format/source/software_process.rst
rly: I changed the text to be consistent with the extension versioning guidelines. Let me know what you think.

oruebel (approved):

bendichter (approved):

PR #425, opened 2020-03-28
add process attribute to TimeSeries.data

### LINKED ISSUE #177: TimeSeries discrete vs continuous vs step
I am working with MWorks data where a sample is stored only when a value changes. I'd like to store these as TimeSeries. I usually think of TimeSeries as values that you could interpolate smoothly between the values. Would it be possible to add a flag for interpolation using previous values? For example

This data:

data = [0, 1, 0, 1]
timestamps = [1, 2, 3, 4]

I would usually upsample using interpolation:

data =  [0, .5, 1, .5, 0, .5, 1]
timestamps = [1, 1.5, 2, 2.5, 3, 3.5, 4]

but I want to indicate that it should be upsampled this way, using the previous value:

data = [0, 0, 1, 1, 0, 0, 1]
timestamps = [1, 1.5, 2, 2.5, 3, 3.5, 4]

bendichter:
fix #177

I wasn't quite sure what to call it. Suggestions welcome.

* [x] schema change
* [x] release notes
* [x] pynwb implementation

### DIFF
--- a/core/nwb.base.yaml
+++ b/core/nwb.base.yaml
@@ -119,7 +119,7 @@ groups:
     - name: continuity
       dtype: text
       doc: Optionally describe the continuity of the data. Can be "continuous", "instantaneous", or
-        "step". For example: A voltage trace would be "continuous", because samples 
+        "step". For example, a voltage trace would be "continuous", because samples 
         are recorded from a continuous process. An array of lick times would be "instantaneous", 
         because the data represents distinct moments in time. Times of image presentations would be  
         "step" because the picture remains the same until the next timepoint. This field is optional, 
--- a/core/nwb.base.yaml
+++ b/core/nwb.base.yaml
@@ -116,6 +116,17 @@ groups:
       doc: Base unit of measurement for working with the data. Actual stored values are
         not necessarily stored in these units. To access the data in these units,
         multiply 'data' by 'conversion'.
+    - name: process
+      dtype: text
+      doc: Optionally describe the process of the data. Can be "discrete", "continuous", or
+        "step-forward". For example: A lick times would be a "discrete" process, because the data
+        represents distict moments in time. A voltage trace would be a "continuous" process,
+        because samples are recorded from a continuous process. Times of image presentations is
+        "step-forward" because the picture remains the same until the next timepoint. This field
+        is optional, but be useful in providing information about the underlying data. It may
+        inform the way this data is interpretted, the way it is visualized, and what analysis
+        methods are applicable.
+      required: false
   - name: starting_time
     dtype: float64
     doc: Timestamp of the first sample in seconds. When timestamps are uniformly
--- a/core/nwb.base.yaml
+++ b/core/nwb.base.yaml
@@ -119,12 +119,12 @@ groups:
     - name: continuity
       dtype: text
       doc: Optionally describe the continuity of the data. Can be "continuous", "instantaneous", or
-        "step". For example: A lick times would be "instantaneous", because the data
-        represents distict moments in time. A voltage trace would be "continuous", because samples
-        are recorded from a continuous process. Times of image presentations is "step" because
-        the picture remains the same until the next timepoint. This field is optional, but is
-        useful in providing information about the underlying data. It may inform the way this data
-        is interpretted, the way it is visualized, and what analysis methods are applicable.
+        "step". For example: A voltage trace would be "continuous", because samples 
+        are recorded from a continuous process. An array of lick times would be "instantaneous", 
+        because the data represents distinct moments in time. Times of image presentations would be  
+        "step" because the picture remains the same until the next timepoint. This field is optional, 
+        but is useful in providing information about the underlying data. It may inform the way this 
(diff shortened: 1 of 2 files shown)


rly:
I also don't know what the best term is here. "Process" could work. What do you think about "continuity"? What we describe as step-forwards seems equivalent to a right-continuous step function.

oruebel:
The term ``process`` sounds strange to me. In particular it is fuzzy in terms of what kind of process this is referring to. If I understand this correctly, this refers to how the data was sampled to turn help the user determine how to interpolate and process the data. How about calling it ``sampling`` or ``sampling_strategy`` instead?

bendichter:
I mean the data is from a continuous vs discrete process. It's about the process of the underlying data, not how we sample it. Continuity also works and may be more intuitive.

oruebel:
In that case I like ``continuity`` better than ``process``. The problem with ``process`` is that even if the data comes from a continuous process, that does not mean that the approach we used to observe that process allows for continuous interpretation of the data. I think ``continuity`` is more explicit in that sense as it refers to the data itself not the process it represents.

rly:
I also wonder whether "discrete" is the best term for event data. "event" may be more intuitive. 

In signal processing terms, event data are analogous to a set of impulses (or delta functions). They have value x at exactly one time. All other values are 0. So "impulse" would also work, but that feels a little too technical and not as intuitive to the average neuroscientist. 

Also, a `continuity` field of `data` having the value `'discrete'` might suggest that the data come from a discrete set of values rather than a continuous range. If we want to use the value `'discrete'`, then naming the attribute `process` might be better and I would prefer it on the `TimeSeries` itself, rather than on `data`, since the time series (consisting of both time and values) is a discrete-time process.

In summary - I favor either:
1) name the attribute `"continuity"` and have the options be continuous, event, and step-forwards
2) name the attribute `'process'`, put it on the `TimeSeries`, and have the options be continuous, discrete, and step-forwards

oruebel:
I would suggest the following 

* name: ``continuity`` 
* allowable values ``continuous``, ``instantaneous``, ``step``

**Reasoning**
* I think we should stick with common math terms to describe this 
* I think the analogous terms to ``continuous`` for  event data would be ``instantaneous`` 
    * I don't like the term ``discrete`` here because it can be misleading in that we have discretely sampled time in all cases anyways
    * The term ``event`` (while intuitive) seems like it may end up being to specific for a particlar case. Also, since this would be part of the base TimeSeries we should use a generic term. 
* I think we should use ``step`` instead of ``step-forwards``. In math a ``step`` function seems to always step forward and if we say ``step-forwards`` then it seems you would also have to include ``step-backwards``, which seems awkward. I would just stick with the convention that a step function steps fowards in time. 
* I prefer the name ``continuity`` because it describes what the data at hand actually is (which I think what we want to describe), whereas, ``process`` refers to what generated the data (which I don't think is what the intent of this is)

rly:
> I would suggest the following
> 
> * name: `continuity`
> * allowable values `continuous`, `instantaneous`, `step`

That sounds good to me.

core/nwb.base.yaml
```diff
@@ -116,16 +116,15 @@ groups:
...
-        is optional, but be useful in providing information about the underlying data. It may
-        inform the way this data is interpretted, the way it is visualized, and what analysis
-        methods are applicable.
+      doc: Optionally describe the continuity of the data. Can be "continuous", "instantaneous", or
+        "step". For example: A lick times would be "instantaneous", because the data
+        represents distict moments in time. A voltage trace would be "continuous", because samples
+        are recorded from a continuous process. Times of image presentations is "step" because
+        the picture remains the same until the next timepoint. This field is optional, but is
+        useful in providing information about the underlying data. It may inform the way this data
+        is interpretted, the way it is visualized, and what analysis methods are applicable.
```
```suggestion
      doc: Optionally describe the continuity of the data. Can be "continuous", "instantaneous", or
        "step". For example: A voltage trace would be "continuous", because samples 
        are recorded from a continuous process. An array of lick times would be "instantaneous", 
        because the data represents distinct moments in time. Times of image presentations would be  
        "step" because the picture remains the same until the next timepoint. This field is optional, 
        but is useful in providing information about the underlying data. It may inform the way this 
        data is interpreted, the way it is visualized, and what analysis methods are applicable.
```

rly: I fixed some typos and reordered the examples to follow the list in the first sentence.

rly (changes requested):

oruebel:
Where appropriate, we should define the default value (or even fix the value) for specific TimeSeries types. We could do this in separate PRs though.

rly (dismissed):

bendichter:
The CI has already started to pay off! There was a colon in the new "continuity" description that was messing up the yaml file.

rly:
Good catch!

rly (dismissed):

rly (dismissed):

rly:
Group decision was to add `enum` field to the schema language, add support for `enum` in HDMF, and add `enum` to the yaml in this PR.

rly (approved):

PR #436, opened 2020-05-19
Add filtering attribute to `ElectricalSeries`

### LINKED ISSUE #435: Add 'filtering' field for ElectricalSeries
The `LFP` neurodata_type docstring says "Filter properties should noted in the `ElectricalSeries` description or comments field." That could be better structured. Each `ElectricalSeries` should have an optional `'filtering'` attribute.

This is related to the Allen Institute storing their neuropixels data. There is a 'filtering' column on the electrodes table for hardware-level filtering, but the raw signals are almost always filtered further with a high-pass filter for spike sorting and a low-pass filter for the LFP. The properties of these filters can influence the power and interpretation of certain analyses.

rly:
Fix #435

### DIFF
--- a/core/nwb.ecephys.yaml
+++ b/core/nwb.ecephys.yaml
@@ -5,6 +5,15 @@ groups:
     The data field is an int or float array storing data in volts. The first
     dimension should always represent time. The second dimension, if present,
     should represent channels.
+  attributes:
+  - name: filtering
+    dtype: text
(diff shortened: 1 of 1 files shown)


bendichter:
I like this idea. This change looks good to me. Next steps for this:

* nwb schema release notes
* adjust pywnb API and create tests in a new branch

rly:
@bendichter @oruebel please review. I have tested this in a PR in PyNWB.

oruebel (approved):

PR #439, opened 2020-05-21
storage_hdf5: Use groups instead of dataset in quantity

t-b:
Broken since 621c2d58 (Updated documentation of the storage,
2017-04-27).


rly (dismissed):

rly (approved):

rly:
The required "build" PR action does not seem to run for older PRs so I am temporarily disabling the requirement, merging the PR, and re-requiring the check.

PR #448, opened 2020-06-01
add waveforms to Units table

### LINKED ISSUE #248: accessing spike waveforms from units table
We need a standard way to store spike waveforms in a way that they are linked with spike times and to their unit identity. Since we are moving to deprecate `Clustering`, we can't rely on using a shared index between `Clustering` and `SpikeEventSeries`, and we will need to find another way to offer this capability.

One solution might be to add a column to `Units` that is called `'waveforms'` and every cell is a number_of_spikes x number_of_electrode x number_of_samples matrix. An issue with this approach, however, is that we would not be able to store unsorted spike waveforms by the same mechanism. We could also allow `SpikeEventSeries` specifically for unsorted spikes, but then we have two different ways of storing spike waveforms, and we are trying to avoid that.

Another approach would be to have the `'waveforms'` column provide references to rows of the `SpikeEventSeries` dataset. An advantage of this is that we would have the spikes only in a single place. One disadvantage is that it might be tricky to work out the electrodes issue. Is there a single `SpikeEventSeries` or multiple, one for each electrode group. If there are multiple, we will need to work out a way for the column to reference different objects between rows. If there is only one, most of the channels will contain zeroes, since spikes are only picked up on a single channel group at a time, and there will be a mismatch between the '`electrodes'` column and the dimensions of the waveform.

bendichter:
fix #248


core/nwb.misc.yaml
```diff
@@ -277,3 +277,33 @@ groups:
       value: volts
       doc: Unit of measurement. This value is fixed to 'volts'.
       required: false
+  - name: waveforms
+    neurodata_type_inc: VectorData
+    dtype: float
```
rly:
```suggestion
    dtype: float32
```
Best to be explicit about the minimum precision

core/nwb.misc.yaml
```diff
@@ -277,3 +277,33 @@ groups:
...
+    attributes:
+      - name: sampling_rate
+        dtype: float32
+        doc: Sampling rate, in hertz.
+        required: false
+      - name: unit
+        dtype: text
+        value: volts
+        doc: Unit of measurement. This value is fixed to 'volts'.
+        required: false
```
rly, line 320: In the `Units` table, each row is a unit. Since each row of 'waveforms' has 'num_spikes' rows, just like the 'spike_times' column, it should be indexed, so there should be a VectorIndex dataset
```suggestion
        required: false
  - name: waveforms_index
    neurodata_type_inc: VectorIndex
    doc: Index into the waveforms dataset.
    quantity: '?'
```

rly (changes requested):

core/nwb.misc.yaml
bendichter: I don't really like this. We are being "explicit" that the **minimum** precision is `float32`, which means this is allowed to be `float32`, or `float64`, in other words, any float. So I think putting `float` is much more clear.

rly: The problem with `float` is that the user needs to know what dtype `float` maps onto. It is not obvious. In Python, `float` means float64, but here we mean float32 or float64. Similarly, with `int` (and `uint`), the user needs to know that `int` means int32 or int64 and NOT int8 or int16. So saying `int32` is clearer. By specifying dtype and number, the user needs to know only that the number specifies the minimum precision, which they have to know anyway, e.g., for cases where we use `uint8`.

rly: We could make an exception for `float` where `float` == `float32` == float32 or float64, and `int` and `uint` would work differently, but I think that would be more confusing. We could in addition remap `int` and `uint` to mean any int and any uint, but I think that would be very confusing.

bendichter: @rly You think it's confusing  to say "int means any type of integer and uint means any type of unsigned integer"? That seems pretty straightforward to me.

rly (dismissed):

rly:
This looks good to me but let's wait on feedback from at least @oruebel and @ajtritt

bendichter:
Our current plan is to implement indexes of indexes.  @ajtritt 👀

rly:
New suggestion to get around different numbers of electrodes for each waveform:
index of indices.

Requires hdmf to support this and hdmf-common-schema VectorIndex to inherit from VectorData instead of Index.

rly:
TODO:
- [x] finish and clean up a brief description of how tables, ragged arrays, doubly ragged arrays, and table row references work in NWB
- [x] add links to the more complete documentation of these types in hdmf-common-schema

rly (approved):

jiumao2:
I wonder have you modified nwb.units.getRow() to get "waveforms" data?
It seems that the "waveforms" data can't be retrieved with nwb.units.getRow(), while the spikes_time can be retrieved normally.

PR #449, opened 2020-06-03
[WIP] Add compatibility statement

rly:
NWB needs a clear guiding statement about how the official APIs support backward compatibility and in what way NWB is maintaining (or not maintaining) compatibility in different schema versions. This comes up again and again and the answer is still not clear.

I wrote this draft based on our current practices, but a few key questions remain. 
1. I know from discussions with @oruebel and @ajtritt that we are considering allowing new requirements in minor releases of NWB as long as the APIs can handle these changes. The APIs would have to use default values or create placeholder objects when reading older files without the new required fields. Are we going to do this? If so, what is the difference between a change like this and a "breaking" change that is reserved for major version releases?
2. In PyNWB, if a file is written with version 2.2.2 and appended to in a later version
of PyNWB that supports version 2.2.4, PyNWB will cache schema 2.2.4 in the file. At this
point, the file should be compliant with 2.2.4 and not necessarily 2.2.2, though this needs considerably more testing. Is this what we want to happen? We should probably hold off on any claims about this until this is well tested in PyNWB.

I think this statement would be best placed in http://nwb.org, but for now, it can be drafted and discussed openly here.

Edit: in retrospect, this is better drafted and discussed in Google Docs. Use this: https://docs.google.com/document/d/1ARNYxoVncrUG0HDQA9y2d5HEUmOKm3Fm6GNJrockDSc/edit?usp=sharing


bendichter (dismissed):

bendichter:
@rly let's add this

docs/format/source/compatibility.rst
```diff
@@ -0,0 +1,70 @@
...
+An NWB file compliant with NWB schema ``x.y`` is generally compliant with later minor versions of the NWB
+schema ``x.(y+1)``. Later minor ``x.(y+1)`` versions of the NWB schema may add new data types, add new
+optional fields, and make required fields optional.
+
+- For example, a new optional data type may be introduced in NWB 2.1 that is not present in NWB 2.0.
+  Because the new data type is optional, the NWB 2.0 file is still compliant with the NWB 2.1 schema.
+
+Later minor ``x.(y+1)`` and bugfix ``x.y.(z+1)`` versions of the NWB schema will avoid adding new required
+fields. Changes that break this rule will generally not be implemented until the next major ``(x+1).0``
+version of NWB. Rare exceptions to this rule are noted in the NWB schema version release notes.
```
oruebel:
```suggestion
version of NWB. Rare exceptions to this rule are noted in the NWB schema version release notes
and the NWB reference APIs will implement appropriate measures to ensure that ``x.y`` files are still
compatible with the APIs.
```

docs/format/source/compatibility.rst
```diff
@@ -0,0 +1,70 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Both the PyNWB and MatNWB APIs implement backward compatibility for **reading and validating** NWB 2.x files. Both
+APIs can read and validate any NWB file compliant with NWB schema 2.0 and later.
+
+- Note: MatNWB reads NWB files based on the cached schema within the file. If an older schema was used to
+  create the file and the schema was not cached, then additional steps [#f1]_ need to be taken to read the file.
+- If PyNWB reads an NWB file with an older, cached version of the NWB schema and then modifies the file, the file
+  will be written with the newer version of the schema. As a precaution, after writing the file, the file should
+  be validated against the newer version of the schema to ensure compliance.
```
oruebel:
```suggestion
- If PyNWB reads an NWB file with an older, cached version of the NWB schema and then modifies the file,
  changes to the file will be written with the newer version of the schema. As a precaution, when writing to
  a file with different version, the file should be validated against the newer version of the schema to ensure 
  compliance. As a best practice, it is recommended to avoid using different versions to write to the same file.
```

oruebel:
> I think this statement would be best placed in http://nwb.org/, but for now, it can be drafted and discussed openly here

@rly I agree that publishing this as a policy on https://www.nwb.org/policies-overview/ is the right place (instead of merging this here). 

Overall, the text looks good to me. I added a couple of minor suggestions to further clarify the open questions that you raised, but I think what is here is a great start and I agree with Ben that we should move forward with it.

rly:
This PR is outdated because we have since published a compatibility policy largely adapted from the text in this PR here: https://www.nwb.org/versioning-and-compatibility/ . That said, I have made the text adjustments that @oruebel suggested above in the new versioning and compatibility policy and updated the policy version string from 0.1.0 to 0.1.1

PR #450, opened 2020-06-12
add to DecompSeries an optional DTR called source_channels

### LINKED ISSUE #447: DecompositionSeries should take an optional DynamicTableRegion
Both Frank and Chang lab are storing `DecompositionSeries` data but not storing the raw `TimeSeries` that is the source. They would like to indicate which electrodes this `DecompositionSeries` came from. To solve this, we should add an optional field to `DecompositionSeries` called "source_channels" takes a `DynamicTableRegion` which can link to electrodes, ROIs, etc.

bendichter:
fix #447

### DIFF
--- a/core/nwb.misc.yaml
+++ b/core/nwb.misc.yaml
@@ -125,6 +125,10 @@ groups:
   - name: metric
     dtype: text
     doc: The metric used, e.g. phase, amplitude, power.
+  - name: source_channels
+    neurodata_type_inc: DynamicTableRegion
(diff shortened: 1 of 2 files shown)


bendichter:
team was on board. We need to make changes to pynwb to allow for this feature

rly:
Note: unlike in Markdown files, rst files use  double backtick, e.g., \``code\`` to indicate code. Single backtick makes the text italic.

rly (approved):

PR #459, opened 2020-08-05
Overhaul IndexSeries and add "order_of_images" to Images

### LINKED ISSUE #458: `IndexSeries` should have the unit 'N/A' and have uint32 data
1) The `unit` attribute is meaningless for the `IndexSeries` type. "The data field stores the index of the frame in the referenced ImageSeries, and timestamps array indicates when that image was displayed."

2) The `data` dataset of `IndexSeries` is int32. It should be uint32 since negative frame indices makes no sense.

rly:
Fix #458


bendichter:
Maybe add `ImageStack` ndtype, where the first dim is n_image, but has no timing info. This could be `VectorData` so that we could put it into a `DynamicTable` and add metadata

rly:
I added an `ImageStack` type and changed `IndexSeries` to use it. Please review. @oruebel @bendichter @ajtritt

core/nwb.base.yaml
```diff
@@ -37,6 +37,21 @@ datasets:
     doc: Description of the image.
     required: false
 
+- neurodata_type_def: ImageStack
+  neurodata_type_inc: VectorData
+  doc: A 1-dimensional ordered collection of images (by reference). The images should be stored in an Images type.
+  dims:
```
ajtritt: I was thinking this would be a 3D column with shape (n_images, n_x_pixels, n_y_pixels).

rly: That would work as well. But that adds the constraint that all images must be the same size, which is usually, but not always the case. We would also have to add the option of a 4D dataset for non-grayscale images.

rly:
TODO: Allow `Images` and `ImageStack` in the nwbfile/stimulus group

rly:
TODO: consider how images are referenced in a table of image presentations (on time, off time, other metadata). Is it possible to reference an Image from a column in a TimeIntervals or other DynamicTable?

core/nwb.base.yaml
rly: On further thought, I agree. We should cover the situation that happens the vast majority of the time. We can always make a new type that is a 1D dataset of references to Image types that may be of different sizes when needed.

core/nwb.base.yaml
```diff
@@ -234,3 +275,27 @@ groups:
...
+    neurodata_type_inc: VectorData
+    dims:
+    - num_images
+    shape:
+    - null
+    dtype:
+      target_type: Image
+      reftype: object
+    doc: Ordered dataset of references to Image objects stored in the parent group.
+    quantity: '?'
```
oruebel, line 258: Does this work in combination with the ImageStacks type. Can you select images with this both from individual images and a stack of images?

rly: No, but does it need to?

I'm not sure why one would want an ordered collection of ImageStack objects and I would rather not add things that have not yet been asked for or have a clear use case.

oruebel: I think I may be misunderstanding the intended use of ImageStack. Can you clarify what the intended use of ``ImageStack`` vs a collection of ``Image`` objects is?

rly: The primary use case is when a user wants to store the times associated with the presentation of a series of stimulus images. This is done in an `IndexSeries`. `IndexSeries/data` contains a list of indices to images stored in either an `ImageStack` or an `Images` object that has `Images/order` set.

An `ImageStack` is to be used when a user wants to store multiple images with the same size. It is similar to a TIFF stack. I think it is an intuitive representation of these data. Storing images in `Images` with the `Images/order` dataset covers a broader use case: when a user wants to store multiple images that do not necessarily have the same size or number of channels in some order. 

We could remove the `ImageStack` and `ImageStacks` types so that there is only one way to store an ordered collection of Images. That seems simpler, and after further discussion, I think @bendichter 
is in favor of this approach. I am now too. We can always add `ImageStack` and `ImageStacks` later on if it is needed.

Using `IndexSeries` linked to `Images`, to find out which image was presented at the first timestamp, the user (or API) would have to do:
```python
image_index = index_series.data[0]
image = index_series.indexed_images.order[image_index]
```
This is a little indirect but can be explained.

Also I think `order` should be renamed `ordered_images`.

bendichter: Yeah, I think `Images` should have an optional `ordered_images` dataset that stores references to the images in that container in order. That will designate the order of the corresponding images so `IndexSeries` can point to the `Images` object and we know which image `IndexSeries` is referring to with each element value

rly:
After discussing this issue with @oruebel, we decided to postpone resolving this until NWB 2.5.0 while we consider the different options and talk with relevant parties.

Another idea is to create an `ImageReferenceSeries` that has a dataset of references to `Image` objects, rather than use indices into an ordered dataset. We would then deprecate `IndexSeries`. This may be simpler.

bendichter:
@rly , @oruebel  can we revisit this?

oruebel:
@bendichter lets revisit this PR during our next sync

core/nwb.base.yaml
```diff
@@ -234,7 +234,10 @@ groups:
 - neurodata_type_def: Images
   neurodata_type_inc: NWBDataInterface
   default_name: Images
-  doc: A collection of images.
+  doc: An collection of images with an optional way to specify the order of the images.
+    For ordered image collections, if the images have different dimensions
+    and number of channels, use this type. If the images have the same dimensions and
+    number of channels, use the ImageStack type.
```
oruebel:
```suggestion
    and number of channels, use this type. If the images are a time series of images with 
    the same dimensions and number of channels, use the ImageSeries type.
```

oruebel: Or should the last sentence refer to ``IndexSeries``? I just made the suggestion because I noticed that it still mentioned ``ImageStack``

rly:
Upon discussion with @oruebel and @bendichter, we decided to implement:
- Allowed ``NWBFile/stimulus/templates`` to contain ``Images`` objects. (#459)
- Added new optional "ordered_images" dataset to ``Images`` that contains an ordered list of object references to
  ``Image`` objects stored in the same ``Images`` object. (#459)
- Overhauled ``IndexSeries`` type (#459):
  - Fixed dtype of ``data`` dataset of ``IndexSeries`` (int32 -> uint32).
  - Updated ``unit`` attribute of ``data`` to have fixed value "N/A".
  - Updated docstrings for the ``conversion`` and ``resolution`` attributes of ``data`` to indicate that
    these fields are not used.
  - Added link to an ``Images`` object, which contains an ordered collection of images.
    Discouraged use of the ``indexed_timeseries`` link to an ``ImageSeries``.

We decided not to implement the `ImageStack` or `ImageStacks` neurodata types because those data can be represented using an ``Images`` type with the `ordered_images` dataset. If there is a use case where storing image stack data as an N-dimensional dataset is more preferred than storing M (N-1)-dimensional datasets in a group, then we can consider adding an `ImageStack` type. 

We could also revamp `IndexSeries` in a different way by creating a new `ImageReferenceSeries` type that has a dataset of references to Image objects, rather than use indices into an ordered dataset of references to datasets in the `Images` group. Those images could reside within the `ImageReferenceSeries` type or in a separate `Images` group. Then we would deprecate `IndexSeries`. We would probably still want to keep the addition of "ordered_images" to the `Images` type. To me, this seems cleaner, though perhaps more disruptive.

i.e.
```yaml
neurodata_type_def: ImageReferenceSeries
neurodata_type_inc: NWBContainer  
# NOTE this is not a TimeSeries because the dtype of data is not numeric and does not have attributes conversion, resolution, offset, unit
doc: A TimeSeries-like container that represents the times at which different images were displayed or acquired. 
datasets:
  - name: data
    dtype:
      target_type: Image
      reftype: object
    dims:
    - num_times
    shape:
    - null
    doc: References to images stored in this object.
  - neurodata_type_inc: Image
    doc: Images stored in this collection.
    quantity: '+'
  - ...  # description, timestamps, etc.
```

As designed in this PR, the API or user must do the following to get the images used in the `IndexSeries` in order:
```python
images_container = nwbfile.stimulus_templates.get("Images")
images = [images_container.ordered_images[x] for x in index_series.data[:]]
```

Whereas with the `ImageReferenceSeries` above, the API or user must do the following:
```python
images = image_reference_series.data[:]
```

@oruebel and @bendichter, what do you think? I am happy to go either way on new `IndexSeries` vs `ImageReferenceSeries`. Just wanted to present this alternative since we have not discussed it previously.

core/nwb.base.yaml
rly: Good catch. I missed that reference

rly: Fixed

core/nwb.base.yaml
```diff
@@ -246,7 +246,7 @@ groups:
   - neurodata_type_inc: Image
     doc: Images stored in this collection.
     quantity: '+'
-  - name: order
+  - name: ordered_images
```
bendichter: how about "order_of_images"?

bendichter:
looks great!

bendichter (approved):

oruebel (approved):
Looks good to me

PR #498, opened 2022-03-05
Improve documentation for some electrodes table columns

rly:
## Summary of changes

The `rel_x`, `rel_y`, `rel_z`, and `reference` columns of the electrodes table are confusing users.
> And the documentation says the following about the reference column:
reference', 'Description of the reference used for this electrode.'
I fail to understand if this is meant to be an arbitrary string that we can use to reference the electrode of it is mean to work in tandem with rel_x, rel_y, rel_z and be the reference point form which they are measured.
Any pointers?

This PR clarifies the documentation string.

- [x] Update the PyNWB documentation accordingly.

## PR checklist for schema changes
- [x] Add release notes for the PR to `docs/format/source/format_release_notes.rst`


bendichter:
This change does not match with my understanding of this field. rel_x, y, and z and position on the probe. This should remain the same even if the probe is rotated in the brain, so it doesn't make sense to put this in terms of the anatomical coordinates of the brain.

rly:
@bendichter That makes sense. I will update the PR. Would it be useful to allow the user to specify an origin point / the reference frame for the rel_x, rel_y, and rel_z positions? I know for kilosort, the `channel_positions.npy` file stores what is basically `rel_x` and `rel_y`, but where the 0 point is is unclear (close to the tip? the connector?) and may not even be useful. Or perhaps we can punt this issue of reference frames until we have a better overall method for specifying spatial coordinates relative to one another.

bendichter:
@rly I think it would be good for us to try to converge with [ProbeInterface](https://probeinterface.readthedocs.io/en/main/) on these types of questions.

bendichter (dismissed):

bendichter (approved):

PR #506, opened 2022-03-22
Added optionality for x,y,z, filtering and impedance

### LINKED ISSUE #505: Remove some required properties from the extracellular electrodes table
This is an issue resulting from the discussion between CatalystNeuro and the core nwb team today at 2022-04-22.
The idea is to reduce the number of required properties as `nwb-conversion-tools` in particular ends up filling them most of the time with default values. The underlying claim is that most of the time the information is not actually available so we are creating default data with not a strong purpose.

Right now the required properties are the following:
- x
- y
- z
- imp 
- location
-  filtering
-  group

The idea is the remove all of them save group.
If I understand correctly, when this works correctly we would add Quantity: 0 or 1 to most of the fields in the following of the schema:
https://nwb-schema.readthedocs.io/en/latest/format.html?highlight=electrodes#groups-general-extracellular-ephys-electrodes

h-mayorquin:
## Summary of changes

This is the first draft of the changes that should fix #505. This is my first PR here so I will welcome any help with it. I am not sure if anything else should be modified. A brief exploration in the repository gave me the idea that everything is propagated automatically from there.

Also, maybe we could modify the doc for `reference` as it seems quite unspecific as we have discussed before:
https://github.com/NeurodataWithoutBorders/nwb-schema/blob/c90e272c132b6f538d08a01b5e893247ded7d7b8/core/nwb.file.yaml#L325-L329

Previous discussion:
https://nwb-users.slack.com/archives/C5XKC14L9/p1646424876546959

EDIT:
It seems that the `reference` issue is already covered in #498.

## PR checklist for schema changes

- [x] Add release notes for the PR to `docs/format/source/format_release_notes.rst`

### DIFF
--- a/core/nwb.file.yaml
+++ b/core/nwb.file.yaml
@@ -275,18 +275,22 @@ groups:
           neurodata_type_inc: VectorData
           dtype: float32
           doc: x coordinate of the channel location in the brain (+x is posterior).
+          quantity: '?'
         - name: y
           neurodata_type_inc: VectorData
           dtype: float32
           doc: y coordinate of the channel location in the brain (+y is inferior).
+          quantity: '?'
         - name: z
           neurodata_type_inc: VectorData
           dtype: float32
           doc: z coordinate of the channel location in the brain (+z is right).
+          quantity: '?'
         - name: imp
           neurodata_type_inc: VectorData
           dtype: float32
           doc: Impedance of the channel, in ohms.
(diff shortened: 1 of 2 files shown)


bendichter:
looks good, @h-mayorquin! Can you implement the corresponding changes in pynwb?

rly (approved):

PR #512, opened 2022-04-20
add cell_id to IntracellularElectrode

### LINKED ISSUE #416: cell ID (for intracellular electrophysiology)
Somewhat related to #413.

The `nwbfile` has the `identifier`, `session_id` and `experiment_description` parameters that help identify a given recording. 
For intracellular recordings where there are multiple cells in a single session (and saved in separate .nwb files), shouldn't there be an additional identifier, like `cell_id` or `unit_id` ?

I guess this id could also go in `IntracellularElectrode`, but it doesn't look like the right place.

bendichter:
## Summary of changes

fix #416

### DIFF
--- a/core/nwb.icephys.yaml
+++ b/core/nwb.icephys.yaml
@@ -201,7 +201,7 @@ groups:
   datasets:
   - name: cell_id
     dtype: text
-    doc: unique ID of cell
+    doc: unique ID of the cell
     quantity: '?'
   - name: description
     dtype: text
--- a/core/nwb.icephys.yaml
+++ b/core/nwb.icephys.yaml
@@ -199,6 +199,10 @@ groups:
   neurodata_type_inc: NWBContainer
   doc: An intracellular electrode and its metadata.
   datasets:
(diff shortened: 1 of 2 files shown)


oruebel (changes requested):
Can you please update the release notes: https://github.com/NeurodataWithoutBorders/nwb-schema/blob/dev/docs/format/source/format_release_notes.rst

Otherwise this looks good to me

core/nwb.icephys.yaml
```diff
@@ -199,6 +199,10 @@ groups:
   neurodata_type_inc: NWBContainer
   doc: An intracellular electrode and its metadata.
   datasets:
+  - name: cell_id
+    dtype: text
+    doc: unique ID of cell
```
oruebel:
```suggestion
    doc: unique ID of the cell
```

oruebel:
@rly it would be good if we could include this in the next release

rly (approved):

oruebel (approved):

Issue #514, opened 2022-04-21
[Documentation]: TimeSeries and units field


CodyCBakerPhD:
Hey, I noticed that the readthedocs view of TimeSeries has `unit` specified three separate times - is this intended?

![image](https://user-images.githubusercontent.com/51133164/164542119-191e3247-b1ab-4c0d-8537-b820ff0481c8.png)
![image](https://user-images.githubusercontent.com/51133164/164542131-d993e726-cc74-4909-8d4e-60c8ec5315db.png)

bendichter:
I think this is unit for data, timestamps, and starting_time, but what appears that the problem is they aren't marked as such

rly:
Ben is correct. I think the dots before "unit" are supposed to depict the hierarchy, but it is definitely not clear. I created an issue in the docutils https://github.com/hdmf-dev/hdmf-docutils/issues/75

oruebel:
> think this is unit for data, timestamps, and starting_time,

Correct. Currently, this is indicated in the table by the number of "." as part of the prefix. The notation used in the tables is described here https://nwb-schema.readthedocs.io/en/latest/format_description.html#notation I.e., the notation in the table looks correct, but there may be better ways to do this.

oruebel:
If I remember correctly, we did this formatting originally because it was consistent with how NWB 1 described the format and because full paths makes the layout very dense. As such, the desire was to have some way to indicate indentation below the previous item in the table. 

BTW, the symbol used to indicate the indentation level is specified here:

https://github.com/NeurodataWithoutBorders/nwb-schema/blob/4cc938f7dd69b6b92c209b084e15aa0c40656b20/docs/format/source/conf_doc_autogen.py#L72-L73

CodyCBakerPhD:
I believe this was resolved with #516, but I'll reopen if I notice it again.

PR #516, opened 2022-05-03
Make nested elements in spec tables in docs more obvious

rly:
## Summary of changes

- Quick resolution for #514 until we have a better solution in hdmf-docutils. See https://github.com/hdmf-dev/hdmf-docutils/issues/75

## Checklist

For all schema changes:

end in "-alpha"), then:


oruebel (changes requested):
Please also update the description of the notation here, i.e., the prefix symbol in the bullet list and in the table should be updated to reflect the updated formatting 

https://github.com/NeurodataWithoutBorders/nwb-schema/blob/46c964b7f60fd1fe82673a34f7262977f93858e9/docs/format/source/format_description.rst?plain=1#L240-L262

oruebel (dismissed):

oruebel (approved):

PR #523, opened 2022-06-13
Add OnePhotonSeries

### LINKED ISSUE #283: DRAFT: Add OnePhotonSeries neurodata_type
Similar to [TwoPhotonSeries](https://pynwb.readthedocs.io/en/stable/pynwb.ophys.html#pynwb.ophys.TwoPhotonSeries) but for handling one photon calcium imaging data from the [Miniscope](http://miniscope.org/index.php/Main_Page).

Details and differences with TwoPhotonSeries need to be fleshed out.

CodyCBakerPhD:
WIP following the discussion of https://github.com/NeurodataWithoutBorders/nwb-schema/issues/283

## Summary of changes

- Add OnePhotonSeries

## Checklist

For all schema changes:

end in "-alpha"), then:

Resolves #283


core/nwb.ophys.yaml
```diff
@@ -1,4 +1,97 @@
...
+    required: true
+  - name: description
+    dtype: text
+    doc: Description of the device (e.g., model, firmware version, processing software version, etc.)
+      as free-form text.
+    required: false
+  - name: manufacturer
+    dtype: text
+    doc: The name of the manufacturer of the device.
+    required: false
```
CodyCBakerPhD: Do I need to re-define these attributes or are they inherited from `neurodata_type_inc: Device`?

oruebel: If they are defined in ``Device`` and you are not changing them, then yes, they are inherited and there is no need to relist them here

CodyCBakerPhD:
Splitting this PR, which introduces the basic `OnePhotonSeries` object that has some extra optional fields the `TwoPhotonSeries` does not, with the rest of the discussion from https://github.com/NeurodataWithoutBorders/nwb-schema/issues/283 which focused on all the types of custom device metadata that might be needed if someone does not specify the manufacturer/model number (in the case of custom builds). I've stashed work on that kind of thing, though: https://github.com/catalystneuro/nwb-schema/tree/custom_optical_imaging_device

In the end, I think adding enhanced support for custom Devices like this is a separate topic from adding support for `OnePhoton` data itself, which is largely just a minor consideration of the differences with the `TwoPhoton` data.

It's also my opinion that going down the road of thoroughly documenting every possible custom device configuration parameter is so open-ended and dynamic that we should either only support free-form text (which is already supported by the `description` of any `Device`) for specifying as many of these details as the user wishes, or encourage the development of extensions for specific sub-types of devices. But I'm open to discussion on this, so I'll bring it up at the next meeting.

CodyCBakerPhD:
@oruebel @rly @bendichter Met with @DaveOC90 today who confirmed this new object captures and describes all the necessary components of their experiments.

PR is now ready for review.

core/nwb.ophys.yaml
```diff
@@ -1,4 +1,50 @@
...
+    dtype: float32
+    doc: Power of the excitation in mW, if known.
+    required: false
+  - name: intensity
+    dtype: float32
+    doc: Intensity of the excitation in mW/mm^2, if known.
+    required: false
+  datasets:
+  - name: field_of_view
+    dtype: float32
```
oruebel: I think it would be useful to add a ``unit`` attribute for ``field_of_view`` with a fixed value of ``meters``. I think this would also be useful to add in ``TwoPhotonSeries`` but since this is a new type I think it makes sense to just add it from the start.

core/nwb.ophys.yaml
```diff
@@ -1,4 +1,50 @@
...
+      w/ the actual data.
+    required: false
+  - name: exposure_time
+    dtype: float32
+    doc: Exposure time of the sample; often the inverse of the frequency.
+    required: false
+  - name: binning
+    dtype: uint8
+    doc: Amount of pixels combined into 'bins'; could be 1, 2, 4, 8, etc.
+    required: false
```
oruebel, line 23: Could you help me understand what the purpose of the ``binning` attribute is and how the value is being used? I think part of my confusion is that I'm not sure what ``combined into 'bins'`` is referring to. Is the term ``bins`` here used to refer to a pixel in the actual data, where the imaging device acquires multiple pixels and then stores some weighted average for the pixel? If that is case, is knowing the number of pixels that were combined sufficient or do you need to know the matrix with the weights?

oruebel:
Overall the schema looks good to me. I just had a couple of minor questions/comments.

core/nwb.ophys.yaml
CodyCBakerPhD: > where the imaging device acquires multiple pixels and then stores some weighted average for the pixel

This is my understanding of it,  though assuming a uniform weight for each pixel, in which case knowing the number of pixels that were combined is sufficient.

It's a property/setting of the microscope as mentioned in [Guo, C., Blair, G. J., Sehgal, M., Jimka, F. N. S., Bellafard, A., Silva, A. J., ... & Aharoni, D. (2021). Miniscope-LFOV: a large field of view, single cell resolution, miniature microscope for wired and wire-free imaging of neural dynamics in freely behaving animals. bioRxiv.](https://www.biorxiv.org/content/biorxiv/early/2021/11/22/2021.11.21.469394.full.pdf)

where they show it can be scaled to increase the sampling rate (temporal resolution), though I can only assume at the cost of spatial resolution. Similarly I'd imagine for a session of fixed length it would also save room in memory/disk.

That said, I don't often see this either altered or if so, recorded in the associated metadata, so I think most cases of this will be a value of `1` for maximum resolution. But it's at least conceptually something that could be annotated (like the power/intensity; it MUST have been a value, property, or controllable setting of the microscope setup, but quite possible the experimenter simply didn't alter it from baseline default or didn't note it).

core/nwb.ophys.yaml
CodyCBakerPhD: Sure

core/nwb.ophys.yaml
oruebel: Thanks for the clarification

oruebel (dismissed):
Looks good to me. We should ideally also get approval from @bendichter and @rly

core/nwb.ophys.yaml
```diff
@@ -1,4 +1,55 @@
...
+    shape:
+    - - 2
+    - - 3
+    doc: Width, height and depth of image or imaged area, in meters.
+    quantity: '?'
+    attributes:
+    - name: unit
+      dtype: text
+      value: meters
+      doc: Unit of measurement for the width, height, and depth of the image or imaged area.
```
bendichter: It seems like this is handled by `ImagingPlane.manifold` and `ImagingPlane.origin_coords`, no?

CodyCBakerPhD: Hmm.. This field was copied over from the `TwoPhotonSeries`, where it's also defined.

Note that `manifold` is deprecated in favor of `origin_coords` + `grid_spacing`: https://github.com/catalystneuro/nwb-schema/blob/add_one_photon_series/core/nwb.ophys.yaml#L262-L263

As for the difference with the `origin_coords` + `grid_spacing` - my best guess here is maybe you could have several `PhotonSeries` corresponding to the same general `ImagingPlane` or `OpticChannels` but maybe the field of view for each series spans only a subset of the entire plane?

If we can remember why it was designed that way for the `TwoPhotonSeries` case maybe it will help make more sense here, but maybe it should also just be a thing defined only for the `ImagingPlane` as well, not sure...

CodyCBakerPhD:
@oruebel @bendichter Are we good to merge this?

core/nwb.ophys.yaml
bendichter: @oruebel, @rly , any ideas about this?

bendichter: we agreed to remove this

CodyCBakerPhD:
Field of view removed

oruebel (dismissed):

rly:
Looks good. Please add and test a PyNWB API for this before we merge. This looks like a relatively simple data type. I wonder whether the `get_class` for this type in PyNWB is sufficient.

rly (dismissed):

bendichter:
@CodyCBakerPhD I believe this would be a minor change, since it does not break backwards compatibility. Could you please update the changelog accordingly?

CodyCBakerPhD:
@bendichter Done, also included the subject attribute addition with it since they are ['new features added in a back-compatible manner'](https://semver.org/)

bendichter (approved):

oruebel (approved):

oruebel:
> one, also included the subject attribute addition

Where is this included here?

CodyCBakerPhD:
> Where is this included here?

In the changelog sectioning: https://github.com/NeurodataWithoutBorders/nwb-schema/pull/523/files#diff-9223a9a6538334955ca2145d2b76549cdebe59e8842243d6fd74518c43ae9016L6-R14

PR #525, opened 2022-08-15
add age reference attribute

### LINKED ISSUE #412: prenatal age
I am dealing with data that has different types of age, including prenatal. I propose adding an optional attribute to the `Subject.age` dataset that is "age_type". This could be populated with the possible values:

* "birth": standard age from birth. This is assumed if the age_type attribute is omitted, but is recommended if there is some ambiguity or if there are other sessions in the dataset that are prenatal.
* "gestational" : duration since the mother's last normal menstrual period
* "fetal": duration since conception
* Are there any other relevant forms of age?

Then we would encourage users to provide age in ISO 8601: `P(n)Y(n)M(n)DT(n)H(n)M(n)S`

e.g. 19 weeks `'P19W'`, 29 years is `'P29Y'`, 103 days would be `'P103D'`.

It would be nice if users could input this information with a `datetime.timedelta` object in pynwb. I don't know of any parser that goes back and forth between this and a python `datetime.timedelta`, but that doesn't sound very hard. That would also allow us to facilitate queries with time comparisons.

**references**
* [prenatal age terminology
](https://americanpregnancy.org/while-pregnant/fetal-development/)
* [ISO 8601 durations](https://en.wikipedia.org/wiki/ISO_8601#Durations)

bendichter:
fix #412


CodyCBakerPhD:
Would this PR also include the lower/upper bounds or ranges?

CodyCBakerPhD:
Also linking: https://github.com/NeurodataWithoutBorders/nwbinspector/issues/209

bendichter:
@CodyCBakerPhD this PR is solely to handle gestational age. I feel those issues are orthogonal so I'm making separate PRs. Also, the DANDI team wants to use "P90D/P100D" syntax to express age ranges which I think will work just as well and will not require changes to the schema (though it certainly will require changes to the NWB inspector and best practices)

CodyCBakerPhD:
> Also, the DANDI team wants to use "P90D/P100D" syntax

Gotcha, I'll have those ready for the session tomorrow then

core/nwb.file.yaml
```diff
@@ -450,6 +450,12 @@ groups:
     dtype: text
     doc: Age of subject. Can be supplied instead of 'date_of_birth'.
     quantity: '?'
+    attributes:
+      - name: reference
+        doc: "Age is with reference to this event. Can be 'birth' or 
+          'gestational'. If reference is omitted, 'birth' is implied."
```
CodyCBakerPhD, line 456: Is it possible to set default values for this at the schema level? I.e., `default: 'birth'` since `required: false`?

bendichter: I was thinking about this. I suppose it is possible. I wanted to make required: false so that this would be backwards compatible. But if you are writing new data perhaps it would be better to make it explicit by having the default behavior be to write "birth" instead of omitting the arg. I'd be fine either way. Thoughts from the team? Should we give this a default value of "birth"?

oruebel: Defining a default value of `birth` makes sense to me

bendichter:
This maps directly to the DANDI schema: https://github.com/dandi/dandi-schema/blob/979eab30ad61afb4275f5900815ff5c9966407eb/dandischema/model_types.py#L604-L626

core/nwb.file.yaml
```diff
@@ -450,6 +450,12 @@ groups:
...
+        dtype: text
+        required: false
```
bendichter, line 458:
```suggestion
        required: false
        default: birth
```

core/nwb.file.yaml
rly: Defining a default value of `birth` makes sense to me too

rly:
This looks fine to me. Please add to the release notes as well.

core/nwb.file.yaml
```diff
@@ -450,6 +450,13 @@ groups:
...
+        default: birth
```
bendichter:
```suggestion
        default_value: birth
```

oruebel (approved):
Looks good to me

docs/format/source/format_release_notes.rst
```diff
@@ -3,6 +3,11 @@
 Release Notes
 =============
 
+2.5.1 (Upcoming)
```
oruebel, line 6: Following version guidelines https://www.nwb.org/versioning-guidelines/ I believe this should be part of ``2.6`` as it adds functionality in a backward compatible way

oruebel (changes requested):
Please also increment the version in the namespace to ``2.6.0-alpha``

 https://github.com/NeurodataWithoutBorders/nwb-schema/blob/b2fcec0b496654a08bafafff8798b567c89026b5/core/nwb.namespace.yaml#L60

PR #545, opened 2023-07-28
Add stimulus_template to IntracellularStimuliTable

iawbrooks:
## Summary of changes

- Added the optional `stimulus_template` column to `IntracellularStimuliTable`. The purpose of this column is to store stimulus template waveforms, i.e. waveforms that were not _recorded_, but were defined as protocols in the recording software.
From the perspective of writing analysis tools, it is much easier to analyze recordings if this noiseless version of the stimulus is available.

## Checklist

For all schema changes:

end in "-alpha"), then:


docs/format/source/format_release_notes.rst
```diff
@@ -10,6 +10,7 @@ Minor changes
 ^^^^^^^^^^^^^
 - Fixed typos in docstrings. (#560)
 - Modified `OptogeneticSeries` to allow 2D data, primarily in extensions of `OptogeneticSeries`. (#564)
+- Added optional ``stimulus_template`` column to ``IntracellularRecordingsTable`` (#545)
```
oruebel:
```suggestion
- Added optional ``stimulus_template`` column to ``IntracellularStimuliTable`` as part of the ``IntracellularRecordingsTable`` (#545)
```

oruebel (dismissed):

oruebel:
@stephprince does this PR require updating the version of the namespace or is that already done on the main branch?

stephprince:
the namespace version is already updated to 2.7.0-alpha on the main branch so should be ok

oruebel:
@rly this is also ready to merge

rly (approved):

PR #559, opened 2023-12-07
propose more stimulus types

CodyCBakerPhD:
## Summary of changes

@oruebel @rly 

From discussion last week, is this the way were allow more generic types to the stimulus area of the file?

## Checklist

For all schema changes:

end in "-alpha"), then:


rly:
Yes. If this is ready, I can write the corresponding pynwb PR, or you can if you want. Let me know!

CodyCBakerPhD:
> Yes. If this is ready, I can write the corresponding pynwb PR, or you can if you want.

If you can take over this it will likely go much faster; I can certainly help write tests and give real life use cases

There's several applications we'd love to be able to utilize this in for current conversions; one for the Allen Institute, there are some visual stimuli presented described using simple `DynamicTable` types; then for holographic stimulation we may be looking into non-TimeSeries based representations more similar to LabeledEvents; I guess there's also other `ndx-events` based representations that are also considered 'stimuli' rather than just naturally observed events

core/nwb.file.yaml
```diff
@@ -131,6 +131,12 @@ groups:
       - neurodata_type_inc: TimeSeries
         doc: TimeSeries objects containing data of presented stimuli.
         quantity: '*'
+      - neurodata_type_inc: DynamicTable
+        doc: DynamicTable objects containing data of presented stimuli.
+        quantity: '*'
+      - neurodata_type_inc: NWBDataInterface
+        doc: Generic NWB data interfaces, usually from an extension, containing data of presented stimuli.
+        quantity: '*'
```
rly:
```suggestion
        quantity: '*'
      # even though TimeSeries is a NWBDataInterface, we do not remove TimeSeries
      # in order to maintain backwards compatibility in the APIs that
      # use the neurodata_type_inc from the schema to set the variable name
```

core/nwb.file.yaml
```diff
@@ -131,6 +131,15 @@ groups:
...
+      # even though TimeSeries is a NWBDataInterface, we do not remove TimeSeries
```
rly:
```suggestion
      # even though TimeSeries is a child type of NWBDataInterface, we do not remove TimeSeries
```

rly (approved):

PR #566, opened 2024-01-17
Update to most recent hdmf-common-schema

mavaylon1:
## Summary of changes

- Update to most recent hdmf-common-schema
- Update dates to 2024
## Checklist

For all schema changes:

end in "-alpha"), then:


rly (dismissed):

rly:
Looks like we had not updated hdmf-common-schema in 3 years, but that hasn't affected anything AFAIK because PyNWB uses HDMF which uses the latest hdmf-common-schema 1.8.0. I'm not sure if this will affect MatNWB @lawrence-mbf

rly (approved):

lawrence-mbf:
@rly The changes won't show until the next version of the schema since we don't use submodules but otherwise the changes look fine.

PR #567, opened 2024-01-19
Boundary for SpatialSeries

### LINKED ISSUE #524: Request: metadata field in `SpatialSeries` for boundary range
It would be nice if there was a clean / easy way to indicate some position metadata in for position data, in particular the boundary range. This would probably be something to add to `SpatialSeries` (or, possibly, in the `Position`) object. 

The use case is if we have possible boundary ranges across x / y / z dimensions, it's nice to store the possible range, to have a size of the map, regardless of the recorded values. We need this for some experiments, and right now we are having to hack this a bit, but adding separate arrays with boundary definitions. 

I chatted on this briefly with @bendichter - adding a simply min/max boundary to each x / y / z only addresses the square / cube case, but it at least addresses this simple / general case (with more complicated map boundaries perhaps needing to be outsourced to an extension). This new field would presumably be optional, but it would be a nice / clear place to be able to read float values for boundary ranges.

mavaylon1:
## Summary of changes

Optional attribute, `bounds`, for SpatialSeries data.

Fix #524 

## Checklist

For all schema changes:

end in "-alpha"), then:


mavaylon1:
Fix #524

mavaylon1:
Review Notes: For the dtype my assumption was to use float32 (per the original issue poster's notes). When I think of a boundary, I imagined this: `[.56, 1.4]`. I don't believe I need to do anything special to support the fact it would be in an list.

core/nwb.behavior.yaml
```diff
@@ -38,6 +38,10 @@ groups:
         is 'meters'. Actual stored values are not necessarily stored in these units.
         To access the data in these units, multiply 'data' by 'conversion' and add 'offset'.
       required: false
+    - name: bounds
+      dtype: float32
+      doc: The boundary range for the data. The units are the same as the units for the data.
+      required: false
```
bendichter: What about multi-dimensional SpatialSeries? For example a single SpatialSeries can often hold x and y position. I think we could add shape constraints, e.g. (None, 2).

rly: I agree - we should add shape constraints. (None, 2) is OK but we can be more specific - (1, 2), (2, 2), or (3, 2). The first dimension should match the length of the second dimension of `data`.

rly:
```suggestion
    - name: bounds
      dtype: float32
      doc: The boundary range (min, max) for each dimension of `data`. The units are the same as the units for the data.
      shape:
      - - 1
        - 2
      - - 2
        - 2
      - - 3
        - 2 
      dims:
      - - x
        - min,max
      - - x,y
        - min,max
      - - x,y,z
        - min,max
      required: false
```

rly: See suggestion below.

docs/format/source/format_release_notes.rst
```diff
@@ -12,6 +12,7 @@ Minor changes
 - Deprecated `ImagingRetinotopy` neurodata type. (#565)
 - Modified `OptogeneticSeries` to allow 2D data, primarily in extensions of `OptogeneticSeries`. (#564)
 - Added optional ``stimulus_template`` column to ``IntracellularStimuliTable`` as part of the ``IntracellularRecordingsTable``. (#545)
+- Added optional `bounds` attribute to `SpatialSeries.data`. (#567)
```
rly:
```suggestion
- Added optional ``bounds`` attribute to ``SpatialSeries.data``. (#567)
```

RST format uses double backtick for code. Otherwise it will be italicized. I messed this up above too.

rly:
Pinging @mavaylon1 about comments above.

rly:
> Fix #524

FYI @mavaylon1 GitHub will only automatically close this issue if the "fix X" syntax is used in the PR description, not a comment, so I recommend that you edit the first post to include that. ([Example where the PR comment did not close the issue](https://github.com/hdmf-dev/hdmf/pull/968#issuecomment-1788066110))

rly:
2.7.0 has been released, so this will have to move to 2.8.0. Please adjust the version strings appropriately as described in the checklist in the first post.

core/nwb.behavior.yaml
mavaylon1: Can you explain the min max part and how that relates to the dimensions? @rly

rly: If the data represent (x, y, z) positions over time, and x has min/max values -100 to 100, y has min/max values -10 to 10, and z has min/max values -1 to 1, then `bounds` would be [(-100, 100), (-10, 10), (-1, 1)], an array with shape (3, 2). Similarly, if data represents N dimensions of positions over time, then `bounds` would be an array with shape (N, 2).

mavaylon1:
> 2.7.0 has been released, so this will have to move to 2.8.0. Please adjust the version strings appropriately as described in the checklist in the first post.

2.7.0 has not been released

rly:
> > 2.7.0 has been released, so this will have to move to 2.8.0. Please adjust the version strings appropriately as described in the checklist in the first post.
> 
> 2.7.0 has not been released

I released it last month while you were away https://github.com/NeurodataWithoutBorders/nwb-schema/releases/tag/2.7.0 to try to get a pynwb release that includes support for it, but I did not manage to get the pynwb side done in time before my vacation. So PyNWB and MatNWB are due for a release that adds 2.7.0 support.

core/nwb.behavior.yaml
```diff
@@ -38,6 +38,24 @@ groups:
         is 'meters'. Actual stored values are not necessarily stored in these units.
         To access the data in these units, multiply 'data' by 'conversion' and add 'offset'.
       required: false
+    - name: bounds
+      dtype: float32
```
rly: `bounds` should be the same dtype as `data`, which is specified above as `numeric`
```suggestion
      dtype: numeric
```

rly:
Can you revert the hdmf-common-schema update for reasons mentioned here: https://github.com/hdmf-dev/hdmf-common-schema/issues/80#issuecomment-1951503777 ?

mavaylon1:
> Can you revert the hdmf-common-schema update for reasons mentioned here: [hdmf-dev/hdmf-common-schema#80 (comment)](https://github.com/hdmf-dev/hdmf-common-schema/issues/80#issuecomment-1951503777) ?

Done. I also added this to the PR checklist.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -6,6 +6,8 @@
 
 For all schema changes:
 - [ ] Add release notes for the PR to `docs/format/source/format_release_notes.rst`.
+- [ ] Have you included the relevant issue number using "Fix #XXX" notation where XXX is the issue number? By including "Fix #XXX" you allow GitHub to close issue #XXX when the PR is merged.
+- [ ] `hdmf-common-schema` needs to point to the latest release and not the latest of the `main ` branch.
```
rly:
```suggestion
- [ ] Make sure that `hdmf-common-schema` points to the latest release and not the latest commit on the `main` branch.
```

rly:
Looks good. Thanks! Out of an abundance of caution, let's not merge this until we implement and test support for this change in a branch/PR of PyNWB. Can you do that @mavaylon1 ?

mavaylon1:
> Looks good. Thanks! Out of an abundance of caution, let's not merge this until we implement and test support for this change in a branch/PR of PyNWB. Can you do that @mavaylon1 ?

Will do.

mavaylon1:
@rly I made a test for this but it's getting weird on pynwb where we are on a branch built from your branch for schema 2.7 in pynwb branch that has nwb-schema pointing to a branch. 

Let's merge this. Especially since our new policy is to point our schemas to the latest release and not the main/dev of the schema repo, meaning this won't mess with pynwb.

rly (approved):

PR #578, opened 2024-06-27
Add dataset to store list of software that produced a file

### LINKED ISSUE #319: An attribute (or dataset?) list software/library which produced that file/dataset etc
I run into situations where it would have been useful to know which tool, and of what version, has produced a given .nwb file.  I have looked into a h5dump of files and into the schema and found no indication that there is any field to store such information ATM.
But such information could be very useful in many use cases. E.g. 
- to troubleshoot detected problems with the file,
- to reproduce some a analysis (if data stored in .nwb is the result of running an analysis script) 
- to guarantee (where possible/needed) that the same version(s) of the tool(s) used to save the file so to not introduce tool/version specific changes.

E.g., in BIDS (neuroimaging world) 

- sidecar files produced by dcm2niix (upon conversion from DICOM) carry its name/version:

http://datasets.datalad.org/dbic/QA/sub-qa/ses-20180910/anat/sub-qa_ses-20180910_acq-MPRAGE_T1w.json

```
  "ConversionSoftware": "dcm2niix",
  "ConversionSoftwareVersion": "v1.0.20171215 (OpenJPEG build) GCC6.3.0",
```
and I have relied on that information already multiple times to assure that I do convert using the same version of dcm2niix in a given dataset.
Note: this field is not standardized (yet) within BIDS itself, but it might be eventually superseded by the below:

- in [BIDS common derivatives enhancement proposal]( https://bids-specification.readthedocs.io/en/common-derivatives/05-derivatives/01-introduction.html#derived-dataset-and-pipeline-description) there is a provision for "PipelineDescription" record which would include "Description" and "Version".  Note: I am yet to resurrect my claim that such information is of importance for BIDS raw datasets as well since they are also not "given by God" and some tool was used to produce them.

I do appreciate the fact that it would be difficult if not impossible to provide a complete and exhaustive provenance record for all software/versions which ever touched a file.  But that is where IMHO *some* information would be better than *none*, and where nwb standard should provision at least some way to store such information in a standard way.

Edit (20210218):
- I think this field/structure (or an additional one) should also incorporate analogous information on extension libraries used to produce this file and embed their namespaces.

stephprince:
## Summary of changes

Fix #319.

Here is a draft proposal for basic provenance information. Given the discussion in #319, I think the goal of the first draft was an easily shareable, approachable representation where the NWB file has an optional field with the names of the software packages and versions used to generate the data in the file. 

After following up offline with @rly , we discussed that this information should likely:
1. be a dataset instead of an attribute since attributes have size limits and these lists could be large depending on the amount of information the user wants to store.
2. be a string dataset of shape (N, 2) instead of a compound data type to allow more flexible modification in the future if we want to provide the option to save additional information. The con of this approach is that the user will not know the labels without looking at the documentation.

Some further considerations might be:
- It was mentioned that this optional field could be added to all objects (not just the nwb file object). If that is preferred, should it be an optional dataset for the `Container` data type in the `hdmf-common-schema`?
- what should the name of this field be? I used `was_generated_by` based on the PROV naming, but something like `software_versions` might be more interpretable. 

Related pynwb changes here: https://github.com/NeurodataWithoutBorders/pynwb/pull/1924

## Checklist

For all schema changes:

end in "-alpha"), then:

### DIFF
--- a/core/nwb.file.yaml
+++ b/core/nwb.file.yaml
@@ -241,11 +241,11 @@ groups:
       - name: file_name
         dtype: text
         doc: Name of script file.
-    - name: generated_by
+    - name: was_generated_by
       dtype: text
-      doc: Software package names and versions used to generate data contained in
-       an NWB File. For each software pacakge or library, include the name of the
-       software in the first column and the version in the second column.
+      doc: Name and version of software package(s) used to generate data contained in
+       this NWB File. For each software package or library, include the name of the
+       software as the first value and the version as the second value.
       dims:
       - num_sources
       - name, version
--- a/core/nwb.file.yaml
+++ b/core/nwb.file.yaml
@@ -241,6 +241,18 @@ groups:
       - name: file_name
         dtype: text
         doc: Name of script file.
+    - name: generated_by
+      dtype: text
+      doc: Software package names and versions used to generate data contained in
+       an NWB File. For each software pacakge or library, include the name of the
+       software in the first column and the version in the second column.
+      dims:
+      - num_sources
+      - name, version
+      shape:
+      - null
+      - 2
+      quantity: '?'
     - name: stimulus
       dtype: text
       doc: Notes about stimuli, such as how and where they were presented.
--- a/docs/format/source/format_release_notes.rst
+++ b/docs/format/source/format_release_notes.rst
@@ -12,7 +12,7 @@ Minor changes
   ``waveform_sd`` columns ragged. This allows for a different number of waveform means/SDs per unit which is useful
   when each unit is associated with a different number of electrodes and there is a waveform mean/SD for each
   electrode and unit. (#576)
-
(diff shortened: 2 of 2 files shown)


t-b:
Sounds like a good idea.

In MIES we have added our own version info since ages a la

image

but having a builtin definition for that is much preferred.

stephprince:
@t-b great, if I'm correctly interpreting the image you shared, I think all of the MIES version info could be mapped to the proposed (name, version) builtin definition?

e.g. something like: 
```python
[('Igor Pro 64bit', '9.0.6.1.56565'),
 ('MIES', 'Release_2.7_20230809-747-g005144'),
 ('Labnotebook', '23'),
 ('HDF5', '1.10.7'),
 ('Sweep Epoch', '9') 
]
```

t-b:
> I think all of the MIES version info could be mapped to the proposed (name, version) builtin definition?

@stephprince Yes exactly.

rly (approved):

PR #608, opened 2025-02-05
Add DeviceModel neurodata type

### LINKED ISSUE #607: Add `DeviceModel` neurodata type
## Problem
We discussed today at the NWB meeting the need in both https://github.com/catalystneuro/ndx-ophys-devices and https://github.com/catalystneuro/ndx-extracellular-channels for a neurodata type that represents the model of a device as opposed to a specific instance of a device model and the configuration of that instance. 

In new NWB schema 2.8.0, we added an optional string attribute to `Device`: `Device.model` to try to handle this in a simple way. However, this could be improved.

Having a separate `DeviceModel` object, with model properties like manufacturer, that is linked to from `Device` has the following advantages:
1. It deduplicates information when there are multiple instances of a device model present in a session (e.g., multiple Neuropixels 1.0 probes were used and have the same manufacturer and locations and sizes of 960 physical contacts)
2. It is useful when combining data across sessions in a relational database (or archive) to say that these devices use the same model. We built this into Spyglass ([ref](https://github.com/LorenFrankLab/spyglass/blob/master/src/spyglass/common/common_device.py#L315)) where there are a handful of probe types with model-level properties but many different instances of a probe type that have different configurations.

## Proposed solution
Create a new `DeviceModel` type that has a `name`, `manufacturer` (string, required), `model_number` (string, required), and `description` (string, optional).
- Unfortunately, we added `Device.model_name` and `Device.model_number` in the 2.8.0 release, perhaps too hastily. I propose we deprecate those in `Device` and recommend users use their equivalents in the new `DeviceModel` type. 
	- Note that PyNWB has not yet released support for 2.8.0 (a release candidate is out though).
	- Note that MatNWB added support for 2.8.0 in `main` but has not yet made a release
- `Device.manufacturer` has been around for some time. I propose we also deprecate that and recommend users use the equivalent in the new `DeviceModel` type.

## An aside: Device identifiers
For cross-session and cross-experiment harmonization of devices (e.g., "find all the NWB files on DANDI that used a Neuropixels 1.0 probe"), _ideally_ everyone uses the same `DeviceModel` name to represent the same real-world device model, e.g., "Neuropixels 1.0" for a Neuropixels 1.0 probe (as opposed to variations like "NP1.0" or "Neuropixels"). And _ideally_ someone would maintain a registry of devices and their unique identifiers, like an [RRID](https://rrid.site/data/source/nlx_144509-1/search), for those devices. Not only would that aid harmonization,  downstream users could use the identifier to access more thorough information about the model on an external website. ProbeInterface contains a limited [library of probe models](https://github.com/SpikeInterface/probeinterface_library) that are identified by the manufacturer and model name (so no change is needed here). We could create and maintain a registry for all NWB device models, but I think that would be a significant undertaking and responsibility, and perhaps a separate discussion.

rly:
## Summary of changes

Fix #607

## Checklist

For all schema changes:

### DIFF
--- a/docs/format/source/format_release_notes.rst
+++ b/docs/format/source/format_release_notes.rst
@@ -10,6 +10,10 @@ Major changes
 ^^^^^^^^^^^^^
 - Deprecated ``Device.model_number``, ``Device.model_name``, ``Device.serial_number``. Use 
   ``Device.model`` link and new ``DeviceModel`` neurodata type instead.
+- Added ``DeviceModel`` neurodata type to represent the model of a device instead of a specific instance of a device.
+  This deduplicates information when a session involves multiple instances of the same device model, and it helps
+  combine data across sessions and experiments when the model is the same.
+  See https://github.com/NeurodataWithoutBorders/nwb-schema/issues/607 for details.
 
 Minor changes
 ^^^^^^^^^^^^^
--- a/docs/format/source/format_release_notes.rst
+++ b/docs/format/source/format_release_notes.rst
@@ -6,6 +6,11 @@ Release Notes
 2.9.0 (Upcoming)
 ----------------
 
+Major changes
+^^^^^^^^^^^^^
+- Deprecated ``Device.model_number``, ``Device.model_name``, ``Device.serial_number``. Use 
+  ``Device.model`` link and new ``DeviceModel`` neurodata type instead.
+
 Minor changes
 ^^^^^^^^^^^^^
 - Fixed typo and removed HTML tag from doc of behavioral neurodata types. (#600)
--- a/core/nwb.device.yaml
+++ b/core/nwb.device.yaml
@@ -41,6 +41,8 @@ groups:
   neurodata_type_inc: NWBContainer
   doc: Model properties of a data acquisition device, e.g., recording system, electrode, microscope.
     This should be extended for specific types of device models to include additional attributes specific to each type.
+    The name of the DeviceModel should be the most common representation of the model name, e.g.,
+    Neuropixels 1.0, V-Probe, Bergamo III.
   attributes:
   - name: manufacturer
     dtype: text
--- a/core/nwb.device.yaml
+++ b/core/nwb.device.yaml
@@ -1,7 +1,8 @@
 groups:
 - neurodata_type_def: Device
   neurodata_type_inc: NWBContainer
-  doc: Metadata about a data acquisition device, e.g., recording system, electrode, microscope.
+  doc: Metadata about a specific instance of a data acquisition device, e.g., recording system, electrode, microscope.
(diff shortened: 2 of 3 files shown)


rly:
In PyNWB, consider allowing the user to pass all devicemodel arguments to device constructor which creates a devicemodel behind the scenes with the same name + " Model" and links to it.

bendichter (dismissed):
Lgtm. @oruebel  what do you think?

docs/format/source/format_release_notes.rst
```diff
@@ -6,6 +6,11 @@ Release Notes
 2.9.0 (Upcoming)
 ----------------
 
+Major changes
+^^^^^^^^^^^^^
+- Deprecated ``Device.model_number``, ``Device.model_name``, ``Device.serial_number``. Use 
+  ``Device.model`` link and new ``DeviceModel`` neurodata type instead.
```
oruebel: I would add a separate bullet point about _"Added `DevideModel` neurodata type to .... "_ and describe what the purpose is Currently this focuses on what was deprecated but it is not clear what `DeviceModel` is and why it was added. We could also include a link to the issue for further details.

oruebel (changes requested):
Just some minor enhancements to the changelog, otherwise this looks good. What is the plan for supporting this in PyNWB.

oruebel:
> In PyNWB, consider allowing the user to pass all devicemodel arguments to device constructor which creates a devicemodel behind the scenes with the same name + " Model" and links to it.

Since the intent seems to be that folks will create extensions of `DeviceModel`, I'm wondering whether that approach may cause confusion. I'm wondering whether it may be simpler to have some factory method instead, e.g., `create_device` on `NWBFile` that would create both the `Device` and `DeviceModel`

docs/format/source/format_release_notes.rst
rly: done

rly:
> > In PyNWB, consider allowing the user to pass all devicemodel arguments to device constructor which creates a devicemodel behind the scenes with the same name + " Model" and links to it.
> 
> Since the intent seems to be that folks will create extensions of `DeviceModel`, I'm wondering whether that approach may cause confusion. I'm wondering whether it may be simpler to have some factory method instead, e.g., `create_device` on `NWBFile` that would create both the `Device` and `DeviceModel`

That makes sense. But factory methods will not work for extensions either if they require additional fields, so users would still have to create the `DeviceModelExtension` and `DeviceExtension` manually.

oruebel (dismissed):

stephprince (dismissed):

stephprince (dismissed):

stephprince (approved):

PR #610, opened 2025-03-02
FrequencyBandsTable

mavaylon1:
## Summary of changes

- ...

## Checklist

For all schema changes:
- [ ] Have you included the relevant issue number using "Fix #XXX" notation where XXX is the issue number? By including "Fix #XXX" you allow GitHub to close issue #XXX when the PR is merged.

end in "-alpha"), then:

### DIFF
--- a/core/nwb.base.yaml
+++ b/core/nwb.base.yaml
@@ -64,11 +64,21 @@ datasets:
 
 - neurodata_type_def: ExternalImage
   neurodata_type_inc: BaseImage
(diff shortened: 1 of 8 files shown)


rly:
Could you please rename this to `FrequencyBandsTable`? "Bands" is very ambiguous.

mavaylon1:
Review Notes:
There is not "Fix open ticket"
This is not the first schema change.

rly (approved):

PR #623, opened 2025-05-08
Use ExternalImage.data instead of file_path

rly:
## Summary of changes

From #604: the new `BaseImage` type is a dataset where the data has no constraints on dtype or shape but is required.  The new `ExternalImage` has an attribute for `file_path` but because it inherits from `BaseImage`, it is also a dataset and therefore has a required data value. @oruebel and I think that `ExternalImage.data` should be used to store the scalar string file path instead of the attribute.

We cannot yet specify that `ExternalImage.data` must be a scalar, so we leave the shape to be Any shape. `shape: scalar` will be supported in the new 2.0 schema language.

## Checklist

For all schema changes:

end in "-alpha"), then:

### DIFF
--- a/core/nwb.base.yaml
+++ b/core/nwb.base.yaml
@@ -64,12 +64,10 @@ datasets:
 
 - neurodata_type_def: ExternalImage
   neurodata_type_inc: BaseImage
-  doc: A type for referencing an external image file.
-  attributes:
-  - name: file_path
-    dtype: text
-    doc: Path or URL to the external image file.
-    required: true
+  doc: A type for referencing an external image file. The single file path or URL to the external image file should
+    be stored in the dataset.
+  dtype: text
+  # shape: scalar  # this will be supported in the NWB schema language 2.0
 
 - neurodata_type_def: ImageReferences
   neurodata_type_inc: NWBData
--- a/core/nwb.base.yaml
+++ b/core/nwb.base.yaml
@@ -65,7 +65,8 @@ datasets:
 - neurodata_type_def: ExternalImage
   neurodata_type_inc: BaseImage
   doc: A type for referencing an external image file. The single file path or URL to the external image file should
-    be stored in the dataset.
+    be stored in the dataset. This type should NOT be used if the image is stored 
+     in another NWB file and that file is linked to this file.
   dtype: text
   # shape: scalar  # this will be supported in the NWB schema language 2.0
 
--- a/docs/format/source/format_release_notes.rst
+++ b/docs/format/source/format_release_notes.rst
@@ -8,7 +8,7 @@ Release Notes
 
 Major changes
 ^^^^^^^^^^^^^
-- Added `BaseImage` and `ExternalImage` as new neurodata types. The first so both `Image` and `ExternalImage` can inherit from it. The second to store external images (#604)
+- Added `BaseImage` and `ExternalImage` as new neurodata types. The first so both `Image` and `ExternalImage` can inherit from it. The second to store external images (#604, #623)
 
 Minor changes
 ^^^^^^^^^^^^^


h-mayorquin:
I see.

> From https://github.com/NeurodataWithoutBorders/nwb-schema/pull/604: the new BaseImage type is a dataset where the data has no constraints on dtype or shape but is required.

Question:
Where in the schema is the constraine that a DataSet must have a required data attribute? I am looking at the `NWBData` here:

https://github.com/NeurodataWithoutBorders/nwb-schema/blob/5105b0082be2b91aac77b0c635efc9e72b8b6a4d/core/nwb.base.yaml#L1-L4

And `Data` in the common schema:

https://github.com/hdmf-dev/hdmf-common-schema/blob/5b4cbb31dbafcff51ca70bf218f464b186568151/common/base.yaml#L2-L4

And I am not sure where, when and how is the constrained declared.

rly:
Because the type is a dataset spec and not a group spec, the type automatically has a required data element as well as `dtype` and `shape` as top-level keys.

rly:
A different way to say this is that a dataset spec defines a data array

h-mayorquin:
Thanks, would it be fair that is enforced by the schema language?

It is just that I don't see anywhere where this is written explicitly:

https://hdmf-schema-language.readthedocs.io/en/latest/description.html#sec-dataset-spec

Not questioning this of course, just wondering if we should better document this.

h-mayorquin (dismissed):
Thanks for moving this forward. I still think that `file_path` is a better name for this but this should work.

I guess another option would to leave data as an empty string and  still use the attribute somehow? but maybe that has problems as well and some redundancy.

rly:
>  would it be fair that is enforced by the schema language?

What do you mean by this?

I agree - we could be more clear in the schema docs. I created https://github.com/hdmf-dev/hdmf-schema-language/issues/46 to track that improvement.

> I guess another option would to leave data as an empty string and still use the attribute somehow? but maybe that has problems as well and some redundancy.

Yes that is an option but it feels a bit redundant / messy.

h-mayorquin:
>> Would it be fair to say that this is enforced by the schema language?

> What do you mean by this?

I might be a bit confused, but I think what I'm trying to say is the following:

Here:
https://github.com/NeurodataWithoutBorders/nwb-schema/blob/5105b0082be2b91aac77b0c635efc9e72b8b6a4d/core/nwb.base.yaml#L29-L33
the schema enforces that there must be a name attribute.

And here:
https://github.com/NeurodataWithoutBorders/nwb-schema/blob/5105b0082be2b91aac77b0c635efc9e72b8b6a4d/core/nwb.ecephys.yaml#L17-L27
the schema enforces that there must be a dataset named data, including its dtype and shape.

So I think my confusion comes from expecting something like this:
https://github.com/NeurodataWithoutBorders/nwb-schema/blob/5105b0082be2b91aac77b0c635efc9e72b8b6a4d/core/nwb.base.yaml#L1C1-L5
to define a data attribute as well. But in this case, data isn’t actually an attribute—it’s just an (implicit?) property that all datasets have. That’s why I said maybe it's something that’s enforced or declared by the schema language itself.

> Yes, that is an option but it feels a bit redundant/messy.

Right, I trust your instinct on this kind of thing.

rly:
I see. Yes, this is enforced or declared by the schema language itself.

In the same way that this yaml within a group spec declares a dataset named "data" with a specific dtype and shape,
https://github.com/NeurodataWithoutBorders/nwb-schema/blob/5105b0082be2b91aac77b0c635efc9e72b8b6a4d/core/nwb.ecephys.yaml#L17-L27

This yaml declares a dataset without a name but with a neurodata type "BaseImage" with default dtype and shape:
```yaml
datasets:
- neurodata_type_def: BaseImage
  neurodata_type_inc: NWBData
  doc: An abstract base type for image data. Parent type for Image and ExternalImage types.
  attributes:
  - name: description
    dtype: text
    doc: Description of the image.
    required: false
```

Dataset data types are not used often (mostly Image and VectorData subtypes). I admit that how they work sometimes confuses me too.

core/nwb.base.yaml
```diff
@@ -64,12 +64,10 @@ datasets:
...
 - neurodata_type_def: ExternalImage
   neurodata_type_inc: BaseImage
-  doc: A type for referencing an external image file.
-  attributes:
-  - name: file_path
-    dtype: text
-    doc: Path or URL to the external image file.
-    required: true
+  doc: A type for referencing an external image file. The single file path or URL to the external image file should
+    be stored in the dataset.
```
stephprince:
```suggestion
  doc: A type for referencing an external image file. The single file path or URL to the external image file should
    be stored in the dataset. This type should NOT be used if the image is stored 
     in another NWB file and that file is linked to this file.
```

I noticed the text I suggested above was included in the `ImageSeries.external_file` description. If I understand correctly, I think the same applies here? It might be helpful to include to clarify whether this type is for external links to other datasets vs. other external file types.

rly: That makes sense. Thanks!

stephprince (approved):

PR #627, opened 2025-05-14
Add ExternalImage.image_mode and image_format

rly:
## Summary of changes

- Follow-up to #604 and $623 to account for comments in https://github.com/NeurodataWithoutBorders/nwb-schema/issues/603#issuecomment-2616464570 following TAB discussion.

I thought about restricting image modes to common ones, but different programming languages (we care primarily about Python Pillow, Matlab, and Javascript) have different support for different modes (and color spaces and bit depths) in different formats.... it gets quite complicated. We could also remove the `image_mode` attribute entirely or make it optional. It is used only as a hint to the user about what is in the data and as an additional way to validate the data.

However, I think restricting image formats to open formats that are widely supported makes sense. We do not want NWB files that link to proprietary or unreadable image file formats.

## Checklist

For all schema changes:

### DIFF
--- a/core/nwb.base.yaml
+++ b/core/nwb.base.yaml
@@ -64,11 +64,21 @@ datasets:
 
 - neurodata_type_def: ExternalImage
   neurodata_type_inc: BaseImage
-  doc: A type for referencing an external image file. The single file path or URL to the external image file should
-    be stored in the dataset. This type should NOT be used if the image is stored 
-     in another NWB file and that file is linked to this file.
+  doc: A type for referencing an external image file. The single file path or URI to the
+    external image file should be stored in the dataset. This type should NOT be used if
+    the image is stored in another NWB file and that file is linked to this file.
   dtype: text
   # shape: scalar  # this will be supported in the NWB schema language 2.0
+  attributes:
+  - name: image_mode
+    dtype: text
+    doc: Image mode (color mode) of the image, e.g., "RGB", "RGBA", "grayscale", and "LA".
+    required: true
+  - name: image_format
+    dtype: text
+    doc: Common name of the image file format. Only widely readable, open file formats are allowed.
+      Allowed values are "PNG", "JPEG", "GIF", and "TIFF".
+    required: true
 
 - neurodata_type_def: ImageReferences
   neurodata_type_inc: NWBData
--- a/core/nwb.base.yaml
+++ b/core/nwb.base.yaml
@@ -77,7 +77,7 @@ datasets:
   - name: image_format
     dtype: text
     doc: Common name of the image file format. Only widely readable, open file formats are allowed.
-      Allowed values are "PNG", "JPEG", "GIF", and "TIFF".
+      Allowed values are "PNG", "JPEG", and "GIF".
     required: true
 
 - neurodata_type_def: ImageReferences
--- a/core/nwb.base.yaml
+++ b/core/nwb.base.yaml
@@ -73,7 +73,7 @@ datasets:
   - name: image_mode
     dtype: text
     doc: Image mode (color mode) of the image, e.g., "RGB", "RGBA", "grayscale", and "LA".
-    required: true
+    required: false
   - name: image_format
     dtype: text
     doc: Common name of the image file format. Only widely readable, open file formats are allowed.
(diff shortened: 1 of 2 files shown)


rly:
Pillow image modes: https://pillow.readthedocs.io/en/stable/handbook/concepts.html#modes and formats https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html

Formats that MATLAB can read: https://www.mathworks.com/help/matlab/ref/imread.html#bq5qu2w_head

core/nwb.base.yaml
```diff
@@ -64,11 +64,21 @@ datasets:
...
+  doc: A type for referencing an external image file. The single file path or URI to the
+    external image file should be stored in the dataset. This type should NOT be used if
+    the image is stored in another NWB file and that file is linked to this file.
   dtype: text
   # shape: scalar  # this will be supported in the NWB schema language 2.0
+  attributes:
+  - name: image_mode
+    dtype: text
+    doc: Image mode (color mode) of the image, e.g., "RGB", "RGBA", "grayscale", and "LA".
+    required: true
```
h-mayorquin: I agree it should be optional, as you suggest. We can provide tools to add this automatically in NeuroConv or similar conversion libraries, but I think it's helpful to spare users the extra step—especially if they don't know how to extract this information themselves.

core/nwb.base.yaml
```diff
@@ -64,11 +64,21 @@ datasets:
...
+  - name: image_format
+    dtype: text
+    doc: Common name of the image file format. Only widely readable, open file formats are allowed.
+      Allowed values are "PNG", "JPEG", "GIF", and "TIFF".
```
h-mayorquin: webp did not convince you?
https://developers.google.com/speed/webp

rly: I did not find easy ways to open webp images in python and matlab. Maybe that's OK but to me, that's a sign that the images are not widely readable yet. I get that these external images will most likely be stimuli presented to the user so maybe they don't need to be readable in common NWB programming languages. Have you encountered real-life experimental data that uses webp to store images?

core/nwb.base.yaml
```suggestion
    required: false
```

rly: I just made it optional

core/nwb.base.yaml
h-mayorquin: Doesn't pillow reads webp?

https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#webp

The matlab limitation seems like a good argument against though.

> Have you encountered real-life experimental data that uses webp to store images?

I have not.

I think I added the format to the discussion because @yarikoptic mentioned it and it seems the format has some advantages in terms of compression but the criteria of widespread available seems like a good one to me.

h-mayorquin:
One concern I have with using TIFF is that it can contain time series data and image stacks (i.e., multiple images in a single file). I think it is difficult to programmatically ensure that an ExternalFile pointing to a TIFF file contains only a single image which is the intended purpose of the object. Meanwhile, there are widely available tools to convert TIFF to other formats with better or the same compression so I don't see a big advantage on allowing it. Given these challenges, it might be better to avoid allowing TIFF altogether. What do you think?

core/nwb.base.yaml
rly: Ah. I missed that Pillow can read webp. Thanks.

We can always add webp later

h-mayorquin: Agree.

rly:
I think that is fine. We can allow TIFF and consider programmatic ways to detect single page/image vs multiple to try to prevent the issue that you raise.

core/nwb.base.yaml
```diff
@@ -64,11 +64,21 @@ datasets:
...
   # shape: scalar  # this will be supported in the NWB schema language 2.0
+  attributes:
+  - name: image_mode
+    dtype: text
+    doc: Image mode (color mode) of the image, e.g., "RGB", "RGBA", "grayscale", and "LA".
+    required: false
+  - name: image_format
+    dtype: text
+    doc: Common name of the image file format. Only widely readable, open file formats are allowed.
+      Allowed values are "PNG", "JPEG", "GIF", and "TIFF".
```
```suggestion
      Allowed values are "PNG", "JPEG", and "GIF".
```

stephprince (approved):

h-mayorquin (approved):
LGTM

PR #629, opened 2025-05-16
Make SpikeEventSeries.timestamps explicitly required

### LINKED ISSUE #622: Perceived(?) ambiguity w.r.t timestamps of SpikeEventSeries
The documentation (doc key) for `timestamps` of `SpikeEventSeries` states that it should be a required property:

https://github.com/NeurodataWithoutBorders/nwb-schema/blob/0c7896b117dd1207a6bad737781d72c88a5421a5/core/nwb.ecephys.yaml#L97-L115

The definition of `timestamps` in `TimeSeries` is here:

https://github.com/NeurodataWithoutBorders/nwb-schema/blob/0c7896b117dd1207a6bad737781d72c88a5421a5/core/nwb.base.yaml#L197-L214

My understanding of inheritance in the specification is the following: If a dataset/attribute/link or a key of these are not defined in a sub-type, it inherits the value from its parent, or the nearest ancestor that defines the basic data type / key.

But that does not seem to be the case here. The only difference from the specification of `timestamps` in the `SpikeEventSeries` and in the `TimeSeries` is that the quantity is missing in `SpikeEventSeries/timestamps`. When the quantity is missing, the default value should be used, which in this case is 1 (required).

However, if the quantity should be determined based on inheritance, the `SpikeEventSeries/timestamps` dataset should be optional. I have tried to find an explanation in the NWB Specification Language documentation for which interpretation to use, but have not found anything. 

In my understanding, the `SpikeEventSeries/timestamps` should look like this:

```yaml
  - name: timestamps
    quantity: 1
```
basically just redefining the quantity, but not re-specifying all the unchanged specification keys

rly:
## Summary of changes

- Fix #622

## Checklist

For all schema changes:

### DIFF
--- a/docs/format/source/format_release_notes.rst
+++ b/docs/format/source/format_release_notes.rst
@@ -12,7 +12,7 @@ Major changes
   can inherit from it. The second to store external images (#604, #623, #627)
 - Changed ``NWBFile.electrodes`` from a generic ``DynamicTable`` with added columns to a new ``ElectrodesTable`` 
   neurodata type that extends ``DynamicTable`` with added columns. (#539, #624)
-- Made SpikeEventSeries.timestamps explicitly required as described in the documentation. (#629)
+- Made ``SpikeEventSeries.timestamps`` explicitly required as described in the documentation. (#629)
 - Changed ``DecompositionSeries.bands`` from a generic ``DynamicTable`` with added columns to a new ``FrequencyBandsTable`` neurodata type that extends ``DynamicTable`` with added columns. (#610)
 
 Minor changes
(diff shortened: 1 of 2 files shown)


docs/format/source/format_release_notes.rst
```diff
@@ -12,6 +12,7 @@ Major changes
   can inherit from it. The second to store external images (#604, #623, #627)
 - Changed ``NWBFile.electrodes`` from a generic ``DynamicTable`` with added columns to a new ``ElectrodesTable`` 
   neurodata type that extends ``DynamicTable`` with added columns. (#539, #624)
+- Made SpikeEventSeries.timestamps explicitly required as described in the documentation. (#629)
```
rly:
```suggestion
- Made ``SpikeEventSeries.timestamps`` explicitly required as described in the documentation. (#629)
```

stephprince (approved):

PR #644, opened 2025-06-25
Reset default value for attrs in IndexSeries

### LINKED ISSUE #641: Suggestion: Use "default_value: null" for unused  fields in IndexSeries
In `IndexSeries`, the data attributes are not used, but the specification only states this in the doc. 

https://github.com/NeurodataWithoutBorders/nwb-schema/blob/764aa372536db61339ec4ae578e0b375b4b54829/core/nwb.image.yaml#L202-L218

An alternative way to ensure the schema more explicitly enforces this is to set `default_value: null` for these attributes, similar to how `value: N/A` is already used for the `unit` attribute:

```yaml
    attributes:
    - name: conversion
      doc: This field is unused by IndexSeries.
      default_value: null
    - name: resolution
      doc: This field is unused by IndexSeries.
      default_value: null
    - name: offset
      default_value: null
      doc: This field is unused by IndexSeries.
    - name: unit
      value: N/A
      doc: This field is unused by IndexSeries and has the value N/A.
```

Edit: Based on discussion, changed original suggestion of specifying `value` to specifying `default_value` instead

rly:
## Summary of changes

- Fix #641

## Checklist

For all schema changes:

### DIFF
--- a/core/nwb.image.yaml
+++ b/core/nwb.image.yaml
@@ -202,14 +202,17 @@ groups:
     attributes:
     - name: conversion
       dtype: float32
+      default_value: null  # reset default value inherited from TimeSeries
       doc: This field is unused by IndexSeries.
       required: false
     - name: resolution
       dtype: float32
+      default_value: null  # reset default value inherited from TimeSeries
       doc: This field is unused by IndexSeries.
       required: false
     - name: offset
       dtype: float32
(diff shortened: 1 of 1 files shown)


rly:
This is proving to be tricky to get right in PyNWB, so I think we should punt this change to the next release.

PR #645, opened 2025-06-26
Implement NWBEP001 (add EventsTable and more)

### LINKED ISSUE #625: Implement NWBEP001
[NWBEP001](https://docs.google.com/document/d/1qcsjyFVX9oI_746RdMoDdmQPu940s0YtDjb1en1Xtdw/edit) has been reviewed and accepted by the TAB after TAB review and community review. [ndx-events](https://github.com/rly/ndx-events/) implements NWBEP001.

We should now adapt and integrate the NWBEP001 schema into the core.

rly:
## Summary of changes

- Fix #625

## Checklist

For all schema changes:

end in "-alpha"), then:


rly:
TODO: Update BehavioralEpochs docstring

core/nwb.event.yaml
```diff
@@ -0,0 +1,133 @@
...
+  doc: A table to store information about the meanings of categorical data. Intended
+    to be used as a lookup table for the meanings of values in a CategoricalVectorData
+    object. All possible values of the parent CategoricalVectorData object should
+    be present in the 'value' column of this table, even if the value is not observed
+    in the data. Additional columns may be added to store additional metadata about
+    each value.
+  datasets:
+  - name: value
+    neurodata_type_inc: VectorData
+    doc: The value of the parent CategoricalVectorData object.
```
rly:
```suggestion
    doc: The value of a row in the parent CategoricalVectorData object.
```

rly:
If I recall correctly, at our team discussion, we decided not to deprecate setting `TimeSeries. continuity="instantaneous"` because 1) it may still be the natural representation in some rare cases (e.g., a regularly sampled signal that happens to represent instantaneous values) and 2) warning/erroring on new in pynwb because it is deprecated breaks existing scripts that set this value, and that cost is not worth the value gained from deprecating continuity="instantaneous"` at the schema level. Also unlike `BehavioralEvents` and `AnnotationSeries`, which are dedicated containers designed for event data storage, `TimeSeries` is a generic base type that has many different uses.

core/nwb.event.yaml
```diff
@@ -0,0 +1,77 @@
+datasets:
+- neurodata_type_def: TimestampVectorData
+  neurodata_type_inc: VectorData
+  dtype: float
+  dims:
+  - num_times
+  shape:
+  - null
+  doc: A 1-dimensional VectorData that stores timestamps in seconds from the
+    session start time.
```
oruebel: Since `TimestampVectorData` is in a table, I think the behavior is different from `TimeSeries.timestamps` in that we don't assume that consecutive timestamps are related (e.g., timestamps may not be sorted and in an EventsTable consecutive rows may be related to different "processes", i.e., different types of events are being recorded in the same table. I think it could be useful to document. To keep it simple, I think we can simply state that timestamps are not required to be sorted. 

```suggestion
  doc: A 1-dimensional VectorData that stores timestamps in seconds from the
    session start time. Timestamp are not required to be sorted in time.
```

oruebel (dismissed):
Looks good to me. I added minor suggestion to clarify the docs. Feel free to accept/reject that suggestion as you deem appropriate.

oruebel (approved):

PR #658, opened 2025-11-21
Add micrometer units clarification to electrode table's localization coordinates 

### LINKED ISSUE #657: Units for coordinates in the electrodes table
Currently in the [schema](https://nwb-schema.readthedocs.io/en/latest/format.html#groups-general-extracellular-ephys-electrodes), the coordinates in the electrodes table do not have any associated units. By convention, neuroconv uses microns for rel x, rel y, and rel z, but the schema would be much clearer if either

A. the table had a units attribute or
B. the documentation specified, for example, "x coordinate in the electrode group _in microns_."

h-mayorquin:
## Summary of changes
Should fix #657

## Checklist

For all schema changes:

end in "-alpha"), then:


rly:
Thanks @h-mayorquin !

rly (approved):

PR #662, opened 2025-12-11
Add optional device link to TimeSeries

### LINKED ISSUE #602: [Feature]: Add optional device linking to `TimeSeries`
This relates to writing data from NIDQ devices in the context of a SpikeGLX recording. [A more complete discussion can be found here](https://github.com/catalystneuro/neuroconv/issues/1154).

To summarize:

Analog channels from NIDQ devices can represent a variety of non-neural signals (e.g., motion, temperature, MEG). It’s unclear how best to store this information. One option is to save every channel as a separate TimeSeries (e.g., TimeSeriesMotionChannel1, TimeSeriesTemperatureChannel2, TimeSeriesMEGChannel3, etc.). This approach allows the TimeSeries description to include all necessary information and also to use other fields such as gains, offsets, and units for better provenance.

However, this approach has a downside: there is currently no formal way to indicate that all of these TimeSeries were recorded from the same device (the NIDQ board). I think it would be good for provenance if we had the option to link a TimeSeries object to a Device.

What do you think? Am I missing something?

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/.github/CODE_OF_CONDUCT.rst)
- [X] Have you checked the [Contributing](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/CONTRIBUTING.rst) document?
- [X] Have you ensured this change was not already [requested](https://github.com/NeurodataWithoutBorders/pynwb/issues)?

rly:
## Summary of changes

- Fix #602 

Requires changes in PyNWB.

## Checklist

For all schema changes:

end in "-alpha"), then:

### DIFF
--- a/core/nwb.base.yaml
+++ b/core/nwb.base.yaml
@@ -249,6 +249,14 @@ groups:
       stored external to the NWB file, in files storing raw data. Once timestamp
       data is calculated, the contents of 'sync' are mostly for archival purposes.
     quantity: '?'
+  links:
+  - name: device
+    target_type: Device
+    doc: Device used to record this time series. This should not be used to represent
+      a relationship between a stimulation time series and the device used for
+      stimulation; that relationship should be represented through other means in the
+      file.
+    quantity: '?'
 
 - neurodata_type_def: ProcessingModule
   neurodata_type_inc: NWBContainer
--- a/docs/format/source/format_release_notes.rst
+++ b/docs/format/source/format_release_notes.rst
@@ -12,6 +12,7 @@ Minor changes
   that are typically used together for analysis, such as spike sorting. (#659)
 - Specified that units for ``ElectrodesTable`` coordinate fields (``x``, ``y``, ``z``, ``rel_x``, ``rel_y``, ``rel_z``)
   should be in microns. (#658)
+- Added optional link from ``TimeSeries`` to ``Device`` to represent the device used to record the time series. (#662)
 
 
 2.9.0 (June 26, 2025)


core/nwb.base.yaml
```diff
@@ -249,6 +249,14 @@ groups:
       stored external to the NWB file, in files storing raw data. Once timestamp
       data is calculated, the contents of 'sync' are mostly for archival purposes.
     quantity: '?'
+  links:
+  - name: device
```
h-mayorquin, line 253:
```suggestion
  - name: acquisition_device
```

Maybe this would make the expected use more clear?

core/nwb.base.yaml
```diff
@@ -249,6 +249,14 @@ groups:
...
+    target_type: Device
+    doc: Device used to record this time series. This should not be used to represent
```
h-mayorquin, line 255:
```suggestion
doc: >
  Teh device that directly acquired the samples stored in this TimeSeries.
  This field captures data acquisition provenance only and must refer to the
  device that digitized or otherwise recorded the signal.

  When multiple hardware components are involved (e.g., probe → headstage →
  acquisition system), this field must refer to the acquisition system itself,
  i.e., the device that digitized the data, not upstream sensors or
  hardware.

  This field must not be used to encode causal, control, or stimulation
  relationships. For example, it must not refer to devices that delivered
  optogenetic or electrical stimulation, triggered task events, or generated
  control signals. Such relationships should be represented elsewhere in the schema.
```

Two sugggestions:

1. Clarify what happens in multi-device acquisition chains like in a typical ecephys setup where we have: probe/head-stage -> DAQ/acquisition box.  I thinks this fit nicely there as we don't have a way of storing the DAQ (the probe/headstage can be encoded in the ElectrodeGroup device). The same or microscopy where the ImagingPlane captures the device but not the acquisition box that was used to acquire the data.

2. More concrete examples of what should not be here.

What do you think?

h-mayorquin (commented):
Great to see this moving forward. Two suggesions/questions on my side.

rly:
See https://github.com/NeurodataWithoutBorders/nwb-schema/pull/690#discussion_r3246137090 . We should also add this to `EventsTable`

Issue #327, opened 2019-11-13
Add precision attribute to Units spike_times column


rly:
From @bendichter 
> SpikeInterface, expects spike times as uint64, corresponding to the indexes of the samples where they occurred. Spikes, which is used by the Giocomo Lab has a similar constraint. In both cases, we need to supply the sampling rate manually when the acquisition is not present, because it is not discernible from the file otherwise. In general, it might be useful to know the temporal resolution with which spikes were found.

See discussion toward the bottom of https://github.com/NeurodataWithoutBorders/nwb-schema/pull/323

rly:
I think the term `resolution` would be more appropriate than `precision`.

PR #666, opened 2026-01-20
Improving documentation for resolution attribute in the `Units` table

h-mayorquin:
The [`resolution`](https://github.com/NeurodataWithoutBorders/nwb-schema/blob/362c0958528868ac323b1d986ecba5050d0cbd36/core/nwb.misc.yaml#L200-L206) attribute on the `spike_times` dataset was added in [PR #328](https://github.com/NeurodataWithoutBorders/nwb-schema/pull/328) to address [issue #327](https://github.com/NeurodataWithoutBorders/nwb-schema/issues/327). The intent was to store the sampling period of the source acquisition data from which spike times were extracted, allowing users to understand the temporal precision of the spike times.

The current documentation describes it as "the smallest possible difference between two spike times" with a note that it is "usually 1 divided by the acquisition sampling rate". This phrasing is indirect and potentially confusing. Since the intent is to store the inverse of the acquisition sampling rate (i.e., the sampling period), the documentation should state this directly rather than describing it in terms of spike time differences. 

After a discussion with @alejoe91 we thought that this could be improved and here is the result. 

## Checklist

For all schema changes:

end in "-alpha"), then:

### DIFF
--- a/core/nwb.misc.yaml
+++ b/core/nwb.misc.yaml
@@ -201,9 +201,9 @@ groups:
       dtype: float64
       doc: The temporal resolution of the spike times, in seconds. This is typically the sampling period (1/sampling_rate)
         of the acquisition data from which spikes were extracted. If the acquisition data was downsampled before
-        spike sorting, the resolution should be the downsampled frequency (lower sampling frequency, larger resolution value). If interpolation
+        spike sorting, the resolution should be based on the downsampled sampling rate (lower sampling rate, larger resolution value). If interpolation
         was applied during spike sorting, then spike times may fall between samples of the acquisition data, so the
-        resolution should be the interpolation step size (higher sampling frequency, smaller resolution value).
+        resolution should be the interpolation step size (higher sampling rate, smaller resolution value).
       required: false
   - name: obs_intervals_index
     neurodata_type_inc: VectorIndex
--- a/core/nwb.misc.yaml
+++ b/core/nwb.misc.yaml
@@ -200,9 +200,10 @@ groups:
     - name: resolution
       dtype: float64
       doc: The temporal resolution of the spike times, in seconds. This is typically the sampling period (1/sampling_rate)
-        of the acquisition data from which spikes were extracted. Notes - (1) If the acquisition data was downsampled before
-        spike sorting, the resolution will be larger (coarser) than the original acquisition sampling period. (2) If
-        interpolation was applied during spike sorting, spike times may fall between samples of the acquisition data.
+        of the acquisition data from which spikes were extracted. If the acquisition data was downsampled before
+        spike sorting, the resolution should be the downsampled frequency (lower sampling frequency, larger resolution value). If interpolation
+        was applied during spike sorting, then spike times may fall between samples of the acquisition data, so the
+        resolution should be the interpolation step size (higher sampling frequency, smaller resolution value).
       required: false
   - name: obs_intervals_index
     neurodata_type_inc: VectorIndex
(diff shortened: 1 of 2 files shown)


bendichter:
lgtm. @oruebel? @rly?

core/nwb.misc.yaml
```diff
@@ -199,10 +199,10 @@ groups:
...
     - name: resolution
       dtype: float64
-      doc: The smallest possible difference between two spike times. Usually 1 divided by the acquisition sampling rate
-        from which spike times were extracted, but could be larger if the acquisition time series was downsampled or
-        smaller if the acquisition time series was smoothed/interpolated and it is possible for the spike time to be
-        between samples.
+      doc: The temporal resolution of the spike times, in seconds. This is typically the sampling period (1/sampling_rate)
+        of the acquisition data from which spikes were extracted. Notes - (1) If the acquisition data was downsampled before
+        spike sorting, the resolution will be larger (coarser) than the original acquisition sampling period. (2) If
+        interpolation was applied during spike sorting, spike times may fall between samples of the acquisition data.
```
rly:
```suggestion
      doc: The temporal resolution of the spike times, in seconds. This is typically the sampling period (1/sampling_rate)
        of the acquisition data from which spikes were extracted. If the acquisition data was downsampled before
        spike sorting, the resolution should be the downsampled period (larger than the original). If interpolation
        was applied during spike sorting, then spike times may fall between samples of the acquisition data, so the
        resolution should be the interpolation step size (smaller than the original).
```

This revision clarifies what to do in case (2). I also think "Notes - (1) ... (2) ..." is a bit awkward. Two plain sentences are fine.

rly (changes requested):
Looks good, but I have a wording suggestion

core/nwb.misc.yaml
h-mayorquin: Thanks, I added this and more clarifications in the last commit.

core/nwb.misc.yaml
```diff
@@ -199,10 +199,11 @@ groups:
...
       dtype: float64
-      doc: The smallest possible difference between two spike times. Usually 1 divided by the acquisition sampling rate
-        from which spike times were extracted, but could be larger if the acquisition time series was downsampled or
-        smaller if the acquisition time series was smoothed/interpolated and it is possible for the spike time to be
-        between samples.
+      doc: The temporal resolution of the spike times, in seconds. This is typically the sampling period (1/sampling_rate)
+        of the acquisition data from which spikes were extracted. If the acquisition data was downsampled before
+        spike sorting, the resolution should be the downsampled frequency (lower sampling frequency, larger resolution value). If interpolation
+        was applied during spike sorting, then spike times may fall between samples of the acquisition data, so the
+        resolution should be the interpolation step size (higher sampling frequency, smaller resolution value).
```
rly:
```suggestion
      doc: The temporal resolution of the spike times, in seconds. This is typically the sampling period (1/sampling_rate)
        of the acquisition data from which spikes were extracted. If the acquisition data was downsampled before
        spike sorting, the resolution should be based on the downsampled sampling rate (lower sampling rate, larger resolution value). If interpolation
        was applied during spike sorting, then spike times may fall between samples of the acquisition data, so the
        resolution should be the interpolation step size (higher sampling rate, smaller resolution value).
```

The resolution is not the downsampled frequency, but 1 / the downsampled frequency. I would also prefer not to add a new term "frequency" when we already have sampling rate. What do you think about this revision?

h-mayorquin: This is better, thanks!

rly (approved):

PR #690, opened 2026-05-15
Recommend /events as single location for EventsTable; add source attribute

### LINKED ISSUE #689: Where to store EventsTable objects in the NWB file?
@oruebel @bendichter and I discussed this today. I'd like to invite @h-mayorquin, @ehennestad, @vislab, and the broader NWB community to weigh in.

[NWBEP001](https://docs.google.com/document/d/1qcsjyFVX9oI_746RdMoDdmQPu940s0YtDjb1en1Xtdw/edit?tab=t.0#heading=h.o40x21ra4na6) proposed storing `EventsTable` objects in a top-level `/events` group, alongside `/intervals`, `acquisition`, `/units`, etc. Based on later feedback from @oruebel (I believe), in #645, I expanded this to three locations:
- `/acquisition` — events from the acquisition system (Neuralynx .nev, Open Ephys event channels, TTL edges).
- `/processing/<module>` — events derived by user processing (filtering, detection, or annotation).
- `/events` — primary experimental event tables ready for downstream analysis.

**That three-way layout was my call, and I no longer think it's right.** The `/events` vs `/processing` boundary is subjective and will produce inconsistent files across labs. I would like to revisit this and reach consensus on an approach before release.

**Option 1**: `/events` only. (Original proposal.) @bendichter said: Events are like time intervals and should be stored at the top-level like `/intervals`. `/acquisition` is for continuous recorded "signals", and discrete events are always processed or annotations.

**Option 2**: `/acquisition` and `/processing`, no `/events`. @oruebel said: Events are like `TimeSeries` and split naturally along acquisition/processing line. `/intervals` is different because intervals are session-level metadata.

**Option 3**: All three. Rejected for the reason above.

(@bendichter, @oruebel - please correct me if I'm misstating)

**I prefer Option 1**, though I disagree with @bendichter's premise that events are never raw. Many acquisition systems (e.g., Neuralynx, Open Ephys, Plexon) write event data (licks, rewards, stimulus times, trial markers, sync pulses) directly as lists of timestamps, which I consider raw. My reasons for Option 1:

1. Having one place to store and discover event data (`/events`) is worth not having the explicit, path-based labeling of raw vs processed events. If desired, such provenance could be stored in the `EventsTable` description or a new attribute like "source". 
2. Events are conceptually and structurally similar to time intervals in `/intervals` and spike times in `/units`, and they would be used in similar ways during analysis. Therefore, they should be organized in a similar way.
3. Users already report confusion that NWB organizes by processing stage rather than data modality ([Pierre et al., 2024](https://www.jneurosci.org/content/44/38/e0381242024), sections "How should different data types be stored?" and "Data access pain points").

Note: `EventsTable` inherits from `DynamicTable`, so the schema permits placement under `/acquisition` and `/processing` regardless. The question is which location the schema recommends.

Please let me know your thoughts by the end of the week. I'd like to release NWB 2.10.0 ASAP. Thanks, and apologies for the compressed timeline.

rly:
## Summary

Closes #689 (pending consensus on this PR).

- **Single recommended location for `EventsTable`.** All `EventsTable` instances for a session belong in the top-level `/events` group of `NWBFile`, regardless of provenance. Replaces the three-way `/acquisition` vs `/processing/<module>` vs `/events` guidance that landed earlier in the 2.10.0 cycle.
- **New optional `source_description` attribute on `EventsTable`** (text, free-form short phrase) records where the events came from. Examples used in docs: `"Acquisition system"`, `"Thresholding of analog signal ANALOG1 at 3 V"`, `"Manual video review"`. The longer `description` attribute still carries the algorithm/channel narrative.
- **One event type per table.** Clarified in the `EventsTable` type docstring, the `/events` group docstring, and the FAQ that each table should hold events of a single type (same per-event metadata columns across all rows). This was stated in NWBEP001 but not previously reflected in the schema docs.
- **Deprecation docstrings updated.** `BehavioralEvents` and `AnnotationSeries` now point at `EventsTable`, name `/events` as the destination, mention the `source_description` attribute, and spell out the per-field migration mapping (including `data` → an event-marker-named column for `BehavioralEvents`, and `data` → `annotation` column for `AnnotationSeries`).

Discussion captured the consensus for Option 1 from #689 (single location, with optional categorical info on the table itself rather than encoded by path). The `source_description` attribute is intentionally free text rather than a controlled vocabulary, because the acquisition/processing distinction is exactly what made the path-based layout inconsistent across labs in the first place.

## Files changed

- `core/nwb.event.yaml` — `source_description` attribute on `EventsTable`; single-type clarification in type docstring.
- `core/nwb.file.yaml` — rewrote `/events` group docstring for the single-location guidance.
- `core/nwb.behavior.yaml`, `core/nwb.misc.yaml` — rewrote `BehavioralEvents` and `AnnotationSeries` deprecation docstrings.
- `docs/format/source/format_description.rst` — FAQ section on routing time-anchored experimental data updated to single-location.
- `docs/format/source/format_release_notes.rst` — combined #689 entry under Major changes (covers both deprecations + the single-location and `source_description` story); separate Minor changes entry covers schema-side additions.

## Test plan

- [x] Schema loads cleanly via `NamespaceCatalog` (`hdmf-common` 1.9.0 + `core` 2.10.0-alpha) in a `pynwb`/`hdmf` dev env.
- [x] Sphinx docs build (`make doc`) — should be checked before merging; FAQ section changes are RST.
- [ ] PyNWB integration: confirm `EventsTable.source_description` round-trips through `add_events_table` / `get_events_table` and that no schema validation regresses.
- [x] Verify CHANGELOG / release-notes entries render correctly under 2.10.0.

## Notes for reviewers

- The schema still permits `EventsTable` under `/acquisition` or in a `ProcessingModule` (because `EventsTable` inherits from `DynamicTable`). This PR does not forbid that placement; it only changes the recommendation. The FAQ no longer mentions alternate placements at all — please flag if you'd prefer to mention this explicitly.
- The `source_description` attribute is free text. If you'd prefer a controlled vocabulary (e.g., `{acquisition, processing, annotation}`), please weigh in — I argued against it in the discussion thread but it's a reasonable alternative.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/nwb.event.yaml
+++ b/core/nwb.event.yaml
@@ -61,16 +61,17 @@ groups:
       different types of events using a strobed or N-bit encoding, then the description
       should describe which channels were used and how the event time is computed,
       e.g., as the rise time of the first bit.
-  - name: source
+  - name: source_description
     dtype: text
-    doc: Optional short phrase identifying where the events came from. This
-      captures provenance that applies to every row in the table. For example,
-      "Acquisition system" for events emitted directly by the acquisition system
-      (e.g., TTL edges or hardware event channels); "Thresholding of analog signal
-      ANALOG1 at 3 V" for events produced by a detection algorithm run on acquired
-      data; or "Manual video review" for events added by a human annotator. Use
-      `description` for the longer narrative of how the event times were computed
-      (channels used, encoding scheme, algorithm parameters, etc.).
+    doc: Optional short text description of where the events came from, applying
+      to every row in the table. For example, "Acquisition system" for events
+      emitted directly by the acquisition system (e.g., TTL edges or hardware
+      event channels); "Thresholding of analog signal ANALOG1 at 3 V" for events
+      produced by a detection algorithm run on acquired data; or "Manual video
+      review" for events added by a human annotator. This is a free-text label of
+      origin only; use `description` for the longer narrative of how the event
+      times were computed (channels used, encoding scheme, algorithm parameters,
+      etc.).
     required: false
   datasets:
   - name: timestamp
--- a/core/nwb.event.yaml
+++ b/core/nwb.event.yaml
@@ -48,6 +48,10 @@ groups:
     is absent, optional, or mixed across rows. Additional columns may be added to
     store metadata about each event, such as the duration of the event. Examples
     include TTL pulses, licks, rewards, stimulus onsets, and detected ripples.
+    Each `EventsTable` should hold events of a single type, so that all rows share
+    the same set of per-event metadata columns. Events of different types (e.g.,
+    licks and stimulus presentations) should be stored in separate `EventsTable`
+    instances.
   attributes:
   - name: description
     dtype: text
@@ -57,6 +61,17 @@ groups:
       different types of events using a strobed or N-bit encoding, then the description
       should describe which channels were used and how the event time is computed,
       e.g., as the rise time of the first bit.
+  - name: source
+    dtype: text
+    doc: Optional short phrase identifying where the events came from. This
+      captures provenance that applies to every row in the table. For example,
+      "Acquisition system" for events emitted directly by the acquisition system
+      (e.g., TTL edges or hardware event channels); "Thresholding of analog signal
+      ANALOG1 at 3 V" for events produced by a detection algorithm run on acquired
+      data; or "Manual video review" for events added by a human annotator. Use
+      `description` for the longer narrative of how the event times were computed
+      (channels used, encoding scheme, algorithm parameters, etc.).
+    required: false
   datasets:
   - name: timestamp
     neurodata_type_inc: TimestampVectorData
--- a/core/nwb.behavior.yaml
+++ b/core/nwb.behavior.yaml
@@ -72,8 +72,8 @@ groups:
     the `timestamp` column, and the `data` field maps to an additional column
     named after the event marker (e.g., `reward_magnitude`, `port_number`); for
     multi-dimensional `data`, use one column per field. Any other per-event
-    metadata becomes additional columns. Use the `source` attribute on the
-    `EventsTable` to record where the events came from (e.g., "Acquisition
+    metadata becomes additional columns. Use the `source_description` attribute
+    on the `EventsTable` to record where the events came from (e.g., "Acquisition
     system", "Thresholding of analog signal ANALOG1 at 3 V", "Manual video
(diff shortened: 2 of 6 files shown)


core/nwb.event.yaml
```diff
@@ -57,6 +61,17 @@ groups:
       different types of events using a strobed or N-bit encoding, then the description
       should describe which channels were used and how the event time is computed,
       e.g., as the rise time of the first bit.
+  - name: source
+    dtype: text
```
oruebel: This mixes a few things. 
- I would change `source` to `source_description` to clarify that this is a textual description of the source, rather than the actual source.

Based on this description, It sounds like we may also want to consider adding:
- Optional attribute "device" with an object reference to the device used for recording the events
- Optional attribute "source" with an object reference linking to the source Container with the data the events were derived from (e.g., an ImageSeries)

oruebel: I think adding the `device` and `source` attributes is optional and could be added later in separate PRs (and/or later releases), if this requires further discussion. However, I would still rename `source` to `source_description` here

oruebel (changes requested):
Overall this looks good to me. I added a suggestion, please take a look.

core/nwb.event.yaml
rly: Thanks. Adding an optional link `device` is pending here: https://github.com/NeurodataWithoutBorders/nwb-schema/pull/662 . We can extend that to include `EventsTable`. Adding an optional link `source` is a good idea. How `source` was used is still ambiguous, but having the link is better than not and we can iterate on a fuller provenance model later. I'll create an issue.

`EventsTable` is a group and therefore supports links. Any reason to use an attribute with an object reference instead of a link?

oruebel: > Any reason to use an attribute with an object reference instead of a link?

Mainly because it is a single, small reference (so a Dataset is not necessary) and because it avoids possible confusion with VectorData datasets that are actual columns of the table. However, either way works. I would make the choice here based on what is most consistent with how other types do this sort of thing.

oruebel: > Adding an optional link `device` is pending here: #662 .

Since that is an additional item and is already pending, I think this PR is good to

oruebel (approved):

oruebel (approved):

